PR #42, opened 2015-02-10
Added ability to add UIViewController based user defined global entries.

maniak-dobrii:
Makes possible to provide custom UIViewController for user defined global entry, which will be pushed on to the FLEX manager's navigation view controller stack instead of explorer view controller. This could be used to add some custom UI for testers, add different switches/notes, for example network stubbing could be toggled from such custom UI.

### DIFF
--- a/Classes/Explorer Toolbar/FLEXManager.h
+++ b/Classes/Explorer Toolbar/FLEXManager.h
@@ -23,9 +23,18 @@
 /// @param entryName The string to be displayed in the cell.
 /// @param objectFutureBlock When you tap on the row, information about the object returned by this block will be displayed.
 /// Passing a block that returns an object allows you to display information about an object whose actual pointer may change at runtime (e.g. +currentUser)
+/// @param forceObjectExplorer Specifies how to interpret object returned from objectFutureBlock. If YES it will create object/class inspector view controller
+/// for any supplied object, while NO will cause any passed UIViewController subclasses to be pushed into manager's navigation view controller.
 /// @note This method must be called from the main thread.
 /// The objectFutureBlock will be invoked from the main thread and may return nil.
 /// @note The passed block will be copied and retain for the duration of the application, you may want to use __weak references.
+- (void)registerGlobalEntryWithName:(NSString *)entryName
+                  objectFutureBlock:(id (^)(void))objectFutureBlock
+                forceObjectExplorer:(BOOL)forceObjectExplorer;
+
+/**
+ *  The same as -registerGlobalEntryWithName:objectFutureBlock:forceObjectExplorer: passing YES for forceObjectExplorer.
+ */
 - (void)registerGlobalEntryWithName:(NSString *)entryName objectFutureBlock:(id(^)(void))objectFutureBlock;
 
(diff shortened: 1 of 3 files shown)


ryanolsonk:
Nice! I like the added feature.

@JaviSoto, do you have any thoughts on the change or the new API?

Classes/Explorer Toolbar/FLEXManager.h
```diff
@@ -23,9 +23,18 @@
 /// @param entryName The string to be displayed in the cell.
 /// @param objectFutureBlock When you tap on the row, information about the object returned by this block will be displayed.
 /// Passing a block that returns an object allows you to display information about an object whose actual pointer may change at runtime (e.g. +currentUser)
+/// @param forceObjectExplorer Specifies how to interpret object returned from objectFutureBlock. If YES it will create object/class inspector view controller
+/// for any supplied object, while NO will cause any passed UIViewController subclasses to be pushed into manager's navigation view controller.
 /// @note This method must be called from the main thread.
 /// The objectFutureBlock will be invoked from the main thread and may return nil.
 /// @note The passed block will be copied and retain for the duration of the application, you may want to use __weak references.
+- (void)registerGlobalEntryWithName:(NSString *)entryName
```
ryanolsonk: What do you think about separating the two API a little? I think something like:

```objc
- (void)registerGlobalEntryWithName:(NSString *)entryName viewControllerFuture:(UIViewController (^)(void))viewControllerFuture;
```

would be a little more clear than the force `BOOL`. The implementation would also be pretty simple since `FLEXGlobalsTableViewControllerEntry` already consists of a name future and a view controller future.

ryanolsonk:
@maniak-dobrii I left some feedback about a small change to the API. Let me know what you think. We also need you to fill you the contributor form prior to merging: https://docs.google.com/forms/d/1gh9y6_i8xFn6pA15PqFeye19VqasuI9-bGp_e0owy74/viewform

Thanks!

Classes/Explorer Toolbar/FLEXManager.h
JaviSoto: +1. It took me a few minutes to understand what this API was about.

JaviSoto: And for that API, does it have to be a view controller future? or we could make it just take the VC?

ryanolsonk: I think we should stick with a view controller future. That way we can create them on demand and drop them when the user backs out. Otherwise we'd be hanging on to a view controller that may never get used for the life of the app.

maniak-dobrii:
@ryanolsonk Yeah, sounds great, I'll update my code. Already submitted contributor form.

maniak-dobrii:
Done, check my updates. Not sure about how to handle nils returned from view controller future block, so, for now it is just asserted. Nothing crashes if you return nil, just a note in the log, but I wouldn't rely on that behaviour, as it may change in future iOS versions.

ryanolsonk:
Great, thanks for the contribution!

PR #44, opened 2015-02-14
Add rename and delete operations to file browser.

drodriguez:
A long press in any file or directory in the file browser will show a menu with options to rename or delete the selected file or directory.

Renames show an alert view with a text field to write the new name for the file or directory.

Deletes show an alert to confirm the deletion of the file.

Some modifications to the main file browser VC have been done to reload the table view sources after the changes.

### DIFF
--- a/Classes/Global State Explorers/FLEXFileBrowserTableViewController.m
+++ b/Classes/Global State Explorers/FLEXFileBrowserTableViewController.m
@@ -7,20 +7,26 @@
 //
 
 #import "FLEXFileBrowserTableViewController.h"
+#import "FLEXFileBrowserFileOperationController.h"
 #import "FLEXUtility.h"
 #import "FLEXWebViewController.h"
 #import "FLEXImagePreviewViewController.h"
 
-@interface FLEXFileBrowserTableViewController ()
+@interface FLEXFileBrowserTableViewCell : UITableViewCell
+@end
+
+@interface FLEXFileBrowserTableViewController () <FLEXFileBrowserFileOperationControllerDelegate>
 
 @property (nonatomic, copy) NSString *path;
 @property (nonatomic, copy) NSArray *childPaths;
+@property (nonatomic, copy) NSString *searchString;
 @property (nonatomic, strong) NSArray *searchPaths;
 @property (nonatomic, strong) NSNumber *recursiveSize;
 @property (nonatomic, strong) NSNumber *searchPathsSize;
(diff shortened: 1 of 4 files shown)


tijoinc:
Sweet!

ryanolsonk:
This is great! Thanks for the contribution. Looking over it, I have some minor feedback that I'll leave inline, but overall it looks like a very clean change.

Classes/Global State Explorers/FLEXFileBrowserTableViewController.m
```diff
@@ -265,4 +300,82 @@ - (void)openFileController:(NSString *)fullPath
...
+    [self.operationQueue cancelAllOperations];
+    FLEXFileBrowserSearchOperation *newOperation = [[FLEXFileBrowserSearchOperation alloc] initWithPath:self.path searchString:self.searchString];
+    newOperation.delegate = self;
+    [self.operationQueue addOperation:newOperation];
+}
+
+@end
+
+
+@implementation FLEXFileBrowserTableViewCell
```
ryanolsonk: What do you think about implementing the actions in `-tableView:performAction:forRowAtIndexPath:` rather than adding the custom cell subclass and responder chain forwarding? It would also cut out the `indexPath` lookup since that is provided directly to the `UITableViewDelegate` method.

ryanolsonk:
Just the one comment that I think would simplify things a little. Let me know what your thoughts are. Otherwise looks good! After merging, I'll do a quick pass to make sure the styling is consistent with the rest of the project.

Classes/Global State Explorers/FLEXFileBrowserTableViewController.m
drodriguez: I will love that to work, but it doesn't. My theory is that Apple only wants us to use that for copy, cut, and paste, and they don't invoke that method unless is one of those three (maybe others, but not the custom one).

I will give it a try, but I don't think I will make it work without the custom cell and playing with the responser chain (the last part is “optional”, but I didn’t want to give the cell the responsibility of deleting/renaming its file).

ryanolsonk: Ah, got it. Looks like `delete:` exists in the `UIResponderStandardEditActions` informal protocol but there's nothing for renaming. Your approach makes total sense now. I'm going to merge and do a quick style pass. Thanks again for the contribution!

ryanolsonk:
Merged manually to address the conflict in the pbxproj file. Thanks!

PR #45, opened 2015-02-14
Add editor for NSDate values.

drodriguez:
The argument input for dates is an UIDatePicker, set to use gregorian
calendar and UTC time zone (the locale is still the current one).

Unfortunately UIDatePicker doesn’t give an option for showing seconds.

Works in the object explorer and the defaults explorer. The changes
around ArgumentInputViewFactory and DefaultEditorVC allows to introspect
the value for its class and show the right editor (otherwise the JSON
editor is used by default).


ryanolsonk:
This change is awesome. Thank you! Merging.

drodriguez:
Thanks!

PR #47, opened 2015-02-25
Fix fileURLOrData passthrough in network observer

judev:
Completion handler may be expecting fileURL so should always pass through the response that the session task gave us.


ryanolsonk:
Oops! Thanks for the fix!

PR #48, opened 2015-03-02
enhance json detect

DaidoujiChen:
Hello @ryanolsonk,

in some of my case, server return the wrong mime type like "text/html" for json data. I think it will be better using a method `isValidJSONData:` to check it. `prettyJSONStringFromData:` handle only the real json data. how do you feel? thank you.


ryanolsonk:
Hey @DaidoujiChen,

Seems like a nice improvement! I was a little concerned that checking `isValidJSONData:` would be slow, but I did some profiling and it looks fine. I left one comment about a check I think we should leave in, but otherwise it looks good!

DaidoujiChen:
Hello @ryanolsonk,

I am rollback the method `prettyJSONStringFromData:`.

ryanolsonk:
:+1: thanks @DaidoujiChen !

DaidoujiChen:
:)

PR #51, opened 2015-03-08
FLEX requires iOS 7 or higher

larrytin:


ryanolsonk:
:+1: thanks for catching that.

PR #54, opened 2015-03-19
Replace spaces in folder name with underscore.

fabiensanglard:
Some build system don't accept space in folder names. This commit just replaces ' ' with '_' where necessary.


ryanolsonk:
Sorry for the delay here. I'm curious - what build system are you working with? And this is super minor, but do you mind going camel case rather than using underscores?

ryanolsonk:
Also, this change breaks the example project. The paths need to be updated there as well.

fabiensanglard:
The build system is Baze.
I have converted the folder names to camelCase and also fixed the Example project.

ryanolsonk:
Thanks! I squashed this and pushed as 0b4e2318143ec0c3bdae14ab7938b9df76ac9683

PR #55, opened 2015-03-24
fix some imports so we don't depend on pch file

mustafa:
this is useful if this is being added to a dynamic framework


ryanolsonk:
:+1: thanks!

Issue #64, opened 2015-05-12
iOS 8.3 endless recursive call on [FLEXExplorerViewController supportedInterfaceOrientations]


simonlinj:
This behaviour happens on real phone and simulator. but only on iOS 8.3.

Steps to reproduce: 
* Enable Flex
* Test an ActionsSheet or AlertView (no crash)
* Test an ActionsSheet or AlertView (crash)

Observations:
[UIApplication sharedApplication] keyWindow] is changed after the first iteration.

```
2015-05-12 14:33:57.028 UICatalog[50167:7492923] keyWindow: <UIWindow: 0x7fbc5171de80; frame = (0 0; 320 568); gestureRecognizers = <NSArray: 0x7fbc5171cd60>; layer = <UIWindowLayer: 0x7fbc5171ca70>>


2015-05-12 14:34:04.365 UICatalog[50167:7492923] keyWindow: <_UIAlertControllerShimPresenterWindow: 0x7fbc517238e0; frame = (0 0; 320 568); opaque = NO; gestureRecognizers = <NSArray: 0x7fbc5171a6f0>; layer = <UIWindowLayer: 0x7fbc51710700>>
```

![screen shot 2015-05-12 at 2 04 17 pm](https://cloud.githubusercontent.com/assets/1506890/7580955/d8956c32-f8af-11e4-9089-36bcf4ef21b0.png)

simonlinj:
I just found out that this issue is a duplicate of #59 sorry!

but c38b90ee60d2ea9131a01b3e1a7014848e0494c9 doesn't solve the crash

ryanolsonk:
Interesting. c38b90e fixed presenting one alert at a time, but I never thought to try presenting two alerts at the same time. I think your proposed change makes sense, but I'd like to dig into the system logic there to better understand the flow. Thanks for reporting the issue and for the pull request!

simonlinj:
actually, its not presenting two alerts at the same time. but presenting one, close it. then open it again.
the weird thing is that it only happens on iOS 8.3

ryanolsonk:
Oh, sorry I misunderstood. Does it reproduce for you in the example app? I just presented & closed several alerts there with FLEX presented and didn't get a crash (8.3 iPhone 6 simulator).

simonlinj:
Hi sorry for the confusion.. The c38b90e solve the AlertView crash but not the ActionSheets crash. both on simulator and real device

ryanolsonk:
[...]

PR #65, opened 2015-05-12
Fix iOS 8.3 failed to get VC for rotation with AlertView/ActionSheet

simonlinj:
A fix for #64


ryanolsonk:
Thanks for the PR! f7c482ceedd539f225733faacb586c49b3247f39 addresses this with a different approach that also covers some related issues, so I'm going to close this one for now.

PR #67, opened 2015-05-17
Fix build errors while building FLEX as framework for usage in Swift project

modnovolyk:
While building FLEX for usage in Swift project i use !use_frameworks option in my podspec. But with this configuration Xcode (6.3.1) can't build FLEX because of following error:
>Include of non-modular header inside framework module 'FLEX.FLEXSystemLogMessage'

screen shot 2015-05-17 at 13 26 08

So i've took out declaration of *logMessageFromASLMessage:* and *asl.h* import into separate private header.
Also, some additional *#pragma* declarations added to suppress *UISearchDisplayController* deprecated warnings.


ryanolsonk:
Thanks for the PR! I think I'd actually prefer to leave the `FLEXSystemLogMessage` header as is and make all headers except `FLEXManager.h` private. That should be the only class that users need to interact with. What do you think about that?

On the issue of the deprecated warnings, we have an existing approach in that file to mute them with a little less pre-processor noise. There's a property on `FLEXFileBrowserTableViewController` called `searchController` which gives access to the search display controller. I'll quickly push a change to mute those warnings. Looks like they slipped in with the rename/delete files feature.

modnovolyk:
I've reviewed `FLEXManager` and other classes a little bit closer and figured out that current architecture does not suppose extending (subclassing) of current implementation. It looks like perfect one-liner. So yes, hiding everything except FLEXManager.h sounds reasonable.
I've updated branch.
Also i've removed extra pre-processor directives and used `searchController ` instead.

ryanolsonk:
Excellent. Thanks for making those changes! Merging.

PR #68, opened 2015-05-17
Fix Network Debugging compatibility with Alamofire

modnovolyk:
Current implementation of swizzled `dataTaskWithRequest:completionHandler:` (and similar methods from `+[FLEXNetworkObserver injectIntoNSURLSessionAsyncDataAndDownloadTaskMethods]`) does not take into consideration case, when sender provides nil completion block (or calls`-dataTaskWithRequest:`).
 It just wraps completion block (even if it is `nil`) and passes to original implementation, preventing firing of events delegation. From `NSURLSessionDataDelegate` documentation:
>Completion handler block are primarily intended as an alternative to using a custom delegate. If you create a task using a method that takes a completion handler block, the delegate methods for response and data delivery are not called.

As a result, enabled Network Debugging causes ignoring of Alamofires completion blocks as its event handling mechanism relies on `NSURLSessions` delegation.

My fix simply calls original `dataTaskWithRequest:completionHandler:` method implementation in case if nil completion block was provided:
```diff
Classes/Network/PonyDebugger/FLEXNetworkObserver.m
 @@ -479,7 +479,10 @@ + (void)injectIntoNSURLSessionAsyncDataAndDownloadTaskMethods
 
             NSURLSessionTask *(^asyncDataOrDownloadSwizzleBlock)(Class, id, NSURLSessionAsyncCompletion) = ^NSURLSessionTask *(Class slf, id argument, NSURLSessionAsyncCompletion completion) {
                 NSURLSessionTask *task = nil;
-                if ([FLEXNetworkObserver isEnabled]) {
+                // If completion block was not provided sender expect to receive delegated methods or does not
+                // interested in callback at all. In this case we should just call original method implementation
+                // with nil completion block.
+                if ([FLEXNetworkObserver isEnabled] && completion) {
                     NSString *requestID = [self nextRequestID];
                     NSString *mechanism = [self mechansimFromClassMethod:selector onClass:class];
                     NSURLSessionAsyncCompletion completionWrapper = [self asyncCompletionWrapperForRequestID:requestID mechanism:mechanism completion:completion];
                     task = ((id(*)(id, SEL, id, id))objc_msgSend)(slf, swizzledSelector, argument, completionWrapper);
                     [self setRequestID:requestID forConnectionOrTask:task];
                 } else {
                     task = ((id(*)(id, SEL, id, id))objc_msgSend)(slf, swizzledSelector, argument, completion);
                 }
                 return task;
             };
```


ryanolsonk:
Very interesting about the NSURLSession behavior. I had no idea. The fix looks good.

PonyDebugger has taken a different approach to sniffing traffic from NSURLSession which I'm liking better: https://github.com/square/PonyDebugger/pull/151
At some point I think we should look at moving to use that approach. The main downside I see is that it won't work on iOS 7.

Thanks for the PR! Merging.

modnovolyk:
I've thought about reducing amount of swizzling code too. I have another idea. I think it is possible to swizzle only getter/setter methods of `delegate` property (both NSURLConnection and NSURLSession) and create proxy object (delegate) that will both log info and forward invocation to original delegate. We can reduce number of `inject` methods in such way.

Another approach could be to use NSURLProtocol mechanism to hook into iOS URL Loading System. Biggest problem of this method that there is no known (to me) way to extend Apples private  _NSHTTPURLProtocol. I am continue researching this topic.

BaiCanLin:
image
😭😭😢

BaiCanLin:
image

Doing so in some specific circumstances or will collapse，block need copy.This will not rush to collapse。
fix: 
as follows
image

PR #69, opened 2015-05-26
Update podspec

jkyin:
Whether it need update podspec to v2.0.3?


ryanolsonk:
:+1: thanks for the reminder!

PR #77, opened 2015-09-04
Remove unused typedef

tijoinc:
Resolves https://github.com/Flipboard/FLEX/issues/73.


ryanolsonk:
Ah, thanks. Not sure why that doesn't show up in the example project. Perhaps it's compiled with different warnings settings than the Flipboard project.

tijoinc:
Yeah, that's probably the case.

tijoinc:
Thanks!

PR #80, opened 2015-09-24
Convenience method to show/hide explorer depending on current isHidden

evliu:
Not required but may be useful since I typically manually do this with my tap gesture handlers. If you feel it's not necessary, feel free to reject this.


ryanolsonk:
Sure! Thanks for the contribution.

evliu:
awesome, will be using `pod 'FLEX',:git => 'https://github.com/Flipboard/FLEX', :commit => 'e9e084e6f15cb8b6e42cfe52581852884973084d'` until the next batch release! (included it here just incase anyone else wants an easy copy/paste)

PR #92, opened 2015-11-30
- (void)toggle***Tool - It is not necessary to 'dismiss' when 'present'

WangHengHeng:
**If always dismiss first before present, then the property `previousKeyWindow` and `previousStatusBarStyle` will be not effective.**

`previousStatusBarStyle` for example:
1. original: white color
image
2. FLEX presented: black color
image
3. FLEX dismissed: still black color(`previousStatusBarStyle` lose effectiveness)
image


ryanolsonk:
Unfortunately the dismiss before present is there for the keyboard shortcuts. For example, you might trigger presentation of the menu while the view hierarchy list is showing. Is there an alternative approach to solving this issue?

WangHengHeng:
@ryanolsonk I notice that when present, the keyWindow changed. so I add a method `dismissPreviousPresentViewController` to solve this issue.

ryanolsonk:
Thanks for contributing this fix! Merging.

PR #104, opened 2016-01-28
Add contributing stub

dcreemer:
move to more permissive contribution doc.
@timonus @ryanolsonk


ryanolsonk:
Thanks for making this more permissive, @dcreemer

PR #105, opened 2016-02-09
Added Realm Database File Introspection Alongside SQLite

TimOliver:
This PR introduces the ability to introspect the contents of Realm database files, using the system that was already put in place to accomodate SQLite files.

Being wary that requiring Realm as an explicit dependency in FLEX would be very cumbersome for any non-Realm users, all of the code has been made optional via the `#if __has_include(<Realm/Realm.h>)` macro. As such, in order for it to be included in the build, it will be necessary for the project to already have the Realm Objective-C framework present at FLEX's build time.

Please let me know if you require anything further in order for this PR to be accepted.

**Full Disclosure:** While I am a full-time employee of Realm, at present we are not in a position to be able to provide official support for Realm integrations in external projects. But I am more than happy to assist if anyone has any queries in my spare time. :)


ryanolsonk:
This is awesome! I'll try to carve out some time to review this weekend. Thanks for the PR!

TimOliver:
Thanks a lot Ryan! 

No worries! If you need any clarification, or need me to make any adjustments to the PR, please let me know! :)

Classes/GlobalStateExplorers/DatabaseBrowser/FLEXDatabaseManager.h
```diff
@@ -13,10 +13,10 @@
 
 #import <Foundation/Foundation.h>
 
-@interface FLEXDatabaseManager : NSObject
+@protocol FLEXDatabaseManager <NSObject>
 
-
-- (instancetype)initWithPath:(NSString*)aPath;
+@optional
```
ryanolsonk: :thumbsup: for making this a protocol rather than subclassing.

Could we make these `required`? Looks like we're not doing `respondsToSelector:` check anyways, and I'd prefer the added compiler safety that comes with `required`.

Classes/GlobalStateExplorers/DatabaseBrowser/FLEXRealmDatabaseManager.m
```diff
@@ -0,0 +1,103 @@
...
+//  FLEXRealmDatabaseManager.m
+//  FLEX
+//
+//  Created by Tim Oliver on 28/01/2016.
+//  Copyright © 2016 Realm. All rights reserved.
+//
+
+#import "FLEXRealmDatabaseManager.h"
+
+#if __has_include(<Realm/Realm.h>)
```
ryanolsonk: :+1:

Classes/GlobalStateExplorers/FLEXFileBrowserTableViewController.m
```diff
@@ -217,9 +217,14 @@ - (void)tableView:(UITableView *)tableView didSelectRowAtIndexPath:(NSIndexPath
                 drillInViewController = [[FLEXWebViewController alloc] initWithText:prettyString];
             } else if ([FLEXWebViewController supportsPathExtension:pathExtension]) {
                 drillInViewController = [[FLEXWebViewController alloc] initWithURL:[NSURL fileURLWithPath:fullPath]];
-            } else if ([[subpath pathExtension] isEqualToString:@"db"]) {
+            } else if ([@[@"db", @"sqlite", @"sqlite3"] containsObject:[subpath pathExtension]]) {
```
ryanolsonk: What do you think about moving this to something like `+[FLEXTableListViewController supportsPathExtension:]`? It would match the pattern used by the web view controller, and it would be nice to keep the supported path extension logic close to `-[FLEXTableListViewController databaseManagerForFileAtPath:]` which is similar.

ryanolsonk:
This is great! Super clean. I left a few suggestions but everything looks good overall.

What do you think about including Realm with the example project so we can demo this feature?

nin9tyfour:
I haven't had a good look at this pull request. But noticed you're using some preprocesser macros to determine if Realm is present. Have you instead considered using some of Objective-C's reflection features to determine if some core Realm classes exist within the loaded class list. You can determine if Realm exists using something like objc_getClass("RLMRealm") or NSClassFromString, if non-nil is returned Realm exists. To me it seems like a more portable solution, especially to those who compile Flex into a dynamic library and use it across multiple projects. Just something to consider anyway.

ryanolsonk:
Great point @nin9tyfour. Making this fully dynamic will definitely make it messier, but I can see the advantage of making it work with realm independent of whether it's built with realm. There are other places in this codebase where I've taken *interesting* approaches to support everything inside a dylib (e.g. the toolbar icons).

I'll defer to what ever @TimOliver thinks is best here.

TimOliver:
Thanks a lot for the feedback @ryanolsonk and sorry for the late reply!

Yeah, I was originally hoping to do it like @nin9tyfour suggested. It would be very cool if the integration of Realm could be fully dynamic so that the functionality was available if Realm itself was detected at runtime.

I initially had a very quick play with it, but I couldn't get it to compile. I've never really played with weak linking a framework to that degree before, so I wasn't sure how much extra work it would actually entail.

I'm certainly happy to try and make it dynamic if you have any tips on how to go about doing it. I was wondering if it would have actually meant needing to import all of Realm's headers into the FLEX project.

Also, yep! I can provide a demo Realm file and class if you'd like some test data for the example app!

In any case, thanks for the feedback! I'll add those code changes and report back here. Time providing, I might also see about having another play with making Realm dynamic too!

TimOliver:
Hi everyone! Sorry for the slight delay in updating.

I had a play with making Realm's integration more dynamic, and I THINK I've got it working now. The FLEX framework can now successfully build with or without the Realm framework being present, and still provide Realm introspection capabilities depending on whether the Realm framework is present at runtime. 

I tested it multiple times by linking the Realm framework to the example app, running, and then removing it again, and it was working perfectly. That being said, since I've never done something like that before, please feel free to check out my changes in case there's a cleaner/easier way to do it.

I also integrated the previously requested changes.

Thanks a lot!

nin9tyfour:
Hi Tim, I'll take a look at this later when I'm at my computer. From what I saw when quickly looking at it, you seem to be quite on point. Great stuff!

ryanolsonk:
Awesome! Thanks again for building this feature!

TimOliver:
Excellent! Thanks a lot for that Ryan! I'm really happy I was able to contribute something to FLEX! :)

PR #107, opened 2016-02-16
*Added the ability to determine a UIView's view controller without selection the view controller's view delegate.

nin9tyfour:
This is more of a convenience thing in most cases. It does become quite useful when debugging parent and child view controllers and is quite helpful (personally) when messing with apps and injecting FLEX using Cydia Substrate. `_viewControllerForAncestor` is quite an old method and has been around since [iOS 4.2 (November 2010)](https://github.com/nst/iOS-Runtime-Headers/blob/4.2/Frameworks/UIKit.framework/UIView.h#L279) so it can be assumed that it is as safe `_viewDelegate` which was introduced in 4.0, I believe. It is, however, more robust in that the view controller can be determined from *any* view contained within a given view controller.

I left the old method as is, even though this new method should be a suitable replacement. I'll leave such changes up to the main contributors' discretion going forward.

### DIFF
--- a/Classes/ObjectExplorers/FLEXViewExplorerViewController.m
+++ b/Classes/ObjectExplorers/FLEXViewExplorerViewController.m
@@ -15,7 +15,8 @@
 
 typedef NS_ENUM(NSUInteger, FLEXViewExplorerRow) {
     FLEXViewExplorerRowViewController,
-    FLEXViewExplorerRowPreview
+    FLEXViewExplorerRowPreview,
+    FLEXViewExplorerRowViewControllerForAncestor
 };
 
 @interface FLEXViewExplorerViewController ()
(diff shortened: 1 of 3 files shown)


ryanolsonk:
Cool! Thanks for the contribution! Sorry for making you deal with the object explorer class hierarchy/customization mess. I've been meaning to refactor that but just haven't found the time.

nin9tyfour:
I actually have a few other additions/changes that I've implemented that I'm not sure whether or not you're interested in. I'm not sure of their usefulness outside of my main focus - development of tweaks for jailbroken iOS, but I guess they could be useful to some. My modifications currently exist in the form of a tweak and swizzle methods (this was simply to allow me to update the git repository without having to worry about resolving conflicts and whatnot). With some work, I can probably adapt them for use in the main project.

Namely:
* Inspecting a class from a string.
* Listing the origin of the class (e.g: UIView originates from UIKit.framework).
* Overview of protocols inherited by classes.The code itself traverses the class hierarchy and lists protocols inherited by each superclass.
* A button to dismiss the keyboard if the explorer window is presented with the keyboard visible.
* A controller for viewing all classes loaded by every loaded image.
* A controller for viewing the names of all methods for all classes in all images.
* A controller for viewing the names of all methods for all classes in a selected image.
* A function for simulating a crash.
* I also modified some code to allow inspection of UIAlertViews and UIActionSheets (or UIAlertControllers). This is achieved by enabling the FLEX window to be just over the level of such controllers and then to be reduced upon presenting a view controller.

I've got these implemented in a way that would probably cause app rejection if left in a published build. If you'd like any of these changes implemented, let me know and I'll see what I can do to hide them using the methods that you've used throughout FLEX.

PR #109, opened 2016-02-17
Added support for password protected databases

revolter:
Currently, the password is set for every database connection, even if some databases are not password protected.

I think that the check would look something like [this](https://github.com/sqlitebrowser/sqlitebrowser/blob/dd92f160125e04e97c1a16fcd05a8407ddb183db/src/sqlitedb.cpp#L207) and I don't know if it's necessary. It depends on the users' use cases, myself for example have only one database, which is encrypted.

### DIFF
--- a/Classes/FLEXManager.h
+++ b/Classes/FLEXManager.h
@@ -49,6 +49,10 @@
 
 #pragma mark - Extensions
 
+/// Default database password is @c nil by default.
+/// Set this to the password you want the databases to open with.
+@property (copy, nonatomic) NSString *defaultDatabasePassword;
+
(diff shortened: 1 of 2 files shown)


revolter:
So it looks like `sqlite3_key` function was not found, and it worked for me because I have SQLCipher included or something. Could it be done with a runtime check if `sqlite3_key` exists in the current scope (project)?

revolter:
Maybe use [this solution](http://stackoverflow.com/a/12518433/865175)?

ryanolsonk:
I'm going to close this out since it doesn't build in the current state. Feel free to reopen if you want to move forward with a different solution.

revolter:
I'll make another PR, hopefully soon

PR #112, opened 2016-02-23
Added handling for nil Realm property values

TimOliver:
Sorry about this! It appears a bug involving Realm nullability slipped through the cracks!

This PR adds a small check to ensure properties return from a Realm object are not nil, and substitutes an `NSNull` instance if it is to ensure the data is still valid when passed to an `NSArray` down the line.


TimOliver:
Thanks a lot for write access @ryanolsonk! I hope doing this was okay! :)

ryanolsonk:
:+1: :shipit:

PR #114, opened 2016-02-25
fix crash when object malloc_size no greater than zero

SoXeon:
when the number of objects changed frequently, may be will  crash that add wrong thing to instances


SoXeon:
for example,before you click the cell of NSCFString, the count of it is 23770, but when you push the view controller, the number of NSCFString changed to 25394, the process of changing will let some instance release，so may be crash

ryanolsonk:
Interesting! The heap objects stuff is inherently unsafe but it seems like this could make it a little better. Thanks for the PR!

PR #128, opened 2016-05-20
Replaced deprecated 'path' property in Realm configuration

TimOliver:
Following the latest trends in Swift, Realm made the decision to deprecate using strings for file paths in its API in favour of `NSURL`. As of the latest version of Realm, 0.103.0, the deprecated `path` property was completely removed.

As reported in https://github.com/Flipboard/FLEX/issues/127, this pull request updates the integration of Realm in FLEX to use the new NSURL value for file paths.


ryanolsonk:
Thanks for the quick fix! :shipit:

PR #134, opened 2016-07-06
Search for a specific class

ThePantsThief:
Added a way to jump to a specific class from the System Libraries tab,
since it is sometimes difficult to find a class if you don’t know which
bundle it’s in.

Also fixed a crash in the second commit. Whoever wrote `[FLEXRuntimeUtility valueForIvar:onObject:]` was checking `strcmp(name, "isa") != 0` when I think they intended to write `strcmp(name, "isa") == 0`, so I corrected that.

### DIFF
--- a/Classes/GlobalStateExplorers/FLEXClassesTableViewController.m
+++ b/Classes/GlobalStateExplorers/FLEXClassesTableViewController.m
@@ -68,7 +68,7 @@
 
 - (void)updateTitle
 {
-    NSString *shortImageName = [[self.binaryImageName componentsSeparatedByString:@"/"] lastObject];
+    NSString *shortImageName = self.binaryImageName.lastPathComponent;
     self.title = [NSString stringWithFormat:@"%@ Classes (%lu)", shortImageName, (unsigned long)[self.filteredClassNames count]];
 }
 
--- a/Classes/GlobalStateExplorers/FLEXLibrariesTableViewController.m
+++ b/Classes/GlobalStateExplorers/FLEXLibrariesTableViewController.m
@@ -9,6 +9,7 @@
 #import "FLEXLibrariesTableViewController.h"
 #import "FLEXUtility.h"
 #import "FLEXClassesTableViewController.h"
+#import "FLEXClassExplorerViewController.h"
 #import <objc/runtime.h>
 
 @interface FLEXLibrariesTableViewController () <UISearchBarDelegate>
@@ -17,6 +18,7 @@
 @property (nonatomic, strong) NSArray *filteredImageNames;
 
 @property (nonatomic, strong) UISearchBar *searchBar;
+@property (nonatomic, strong) Class foundClass;
 
 @end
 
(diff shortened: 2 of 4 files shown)


Classes/GlobalStateExplorers/FLEXClassesTableViewController.m
```diff
@@ -68,7 +68,7 @@ - (void)loadClassNames
 
 - (void)updateTitle
 {
-    NSString *shortImageName = [[self.binaryImageName componentsSeparatedByString:@"/"] lastObject];
+    NSString *shortImageName = self.binaryImageName.lastPathComponent;
```
ryanolsonk: 👍

Classes/GlobalStateExplorers/FLEXLibrariesTableViewController.m
```diff
@@ -63,9 +65,9 @@ - (void)loadImageNames
         
         // Sort alphabetically
         self.imageNames = [imageNameStrings sortedArrayWithOptions:0 usingComparator:^NSComparisonResult(NSString *name1, NSString *name2) {
-                NSString *shortName1 = [self shortNameForImageName:name1];
-                NSString *shortName2 = [self shortNameForImageName:name2];
-                return [shortName1 caseInsensitiveCompare:shortName2];
+            NSString *shortName1 = [self shortNameForImageName:name1];
```
ryanolsonk: 👍

Classes/GlobalStateExplorers/FLEXLibrariesTableViewController.m
```diff
@@ -127,22 +129,32 @@ - (NSInteger)numberOfSectionsInTableView:(UITableView *)tableView
...
         cell.textLabel.font = [FLEXUtility defaultTableViewCellLabelFont];
     }
     
-    NSString *fullImageName = self.filteredImageNames[indexPath.row];
-    cell.textLabel.text = [self shortNameForImageName:fullImageName];
+    NSString *executablePath;
+    if (self.foundClass) {
+        if (indexPath.row == 0) {
+            cell.textLabel.text = [NSString stringWithFormat:@"Class \"%@\"", self.searchBar.text];
+            return cell;
```
ryanolsonk: Would prefer to avoid the middle return here. Not a huge deal though.

Classes/Utility/FLEXRuntimeUtility.m
```diff
@@ -209,7 +209,7 @@ + (id)valueForIvar:(Ivar)ivar onObject:(id)object
 #ifdef __arm64__
     // See http://www.sealiesoftware.com/blog/archive/2013/09/24/objc_explain_Non-pointer_isa.html
     const char *name = ivar_getName(ivar);
-    if (type[0] == @encode(Class)[0] && strcmp(name, "isa") != 0) {
+    if (type[0] == @encode(Class)[0] && strcmp(name, "isa") == 0) {
```
ryanolsonk: Nice catch

Classes/GlobalStateExplorers/FLEXLibrariesTableViewController.m
ThePantsThief: I can rework that before you pull if you'd like

ryanolsonk:
Thanks for the contribution!

"Whoever wrote `FLEXRuntimeUtility`..." -- I think that would be me 😉 . Thanks for fixing the bug.

ThePantsThief:
Sure thing! I was browsing commits and the bug wasn't there originally, so I was thinking someone else might have introduced it via a merge or something.

PR #142, opened 2016-08-15
Fix typo pruged -> pruned

shepting:


samdozor:
I assume the original author meant "purged," not "pruned."

ryanolsonk:
I did mean purged, but pruned is fine too :)

Sorry for the latency here.

PR #143, opened 2016-08-21
Fix NSURLSessionDataDelegate optional method issue

huajiahen:
FLEX hook NSURLSessionDataDelegate optional method `-URLSession:dataTask:didReceiveResponse:completionHandler:`. In Apple's document,  if delegate implement this method, they have to call completionHandler, or system will not proceed the request.
To reproduce this bug, remove this function in `Example/UICatalog/AAPLAppDelegate.m`. Then the NSURLSessionDataTask with URL `http://cdn.flipboard.com/serviceIcons/v2/social-icon-flipboard-96.png` will be hang-up forever.
So when inject to this function, we should first check if delegate class has implement this function. If not implemented, we have to call completionHandler every time this method get called.


dcirne:
I too have been affected by this issue. It would be tremendously helpful if this PR gets accepted and merged.

ryanolsonk:
Thanks for the PR! Sorry for the latency on review here. I think #151 has a slightly cleaner approach to fixing this, since we already have a mechanism for providing a different block when the method is not implemented. I've merged that one, which should resolve the issue!

PR #151, opened 2016-09-29
Fix method swizzling for NSURLSessionDelegate

Blankdlh:
`URLSession:task:willPerformHTTPRedirection:newRequest:completionHandler:`
`URLSession:dataTask:didReceiveResponse:completionHandler:`

The completion handler must be invoked If these delegates are implemented. Otherwise, the task will never finish.


ryanolsonk:
Thanks for the fix! LGTM

PR #157, opened 2016-10-20
[FLEX] Add size to heap objects view

froody:


ryanolsonk:
Very cool! Thanks for the contribution.

Issue #131, opened 2016-06-14
How to exclude FLEX.framework from release builds?


coppercash:
Hi,
I have read the section **Excluding FLEX from Release (App Store) Builds** in the README, which suggests to exclude source files by name.
However, I am using **FLEX.framework** with the form of **Embedded Binaries**.
Is there any way to exclude **FLEX.framework** from release builds?
Thanks

ryanolsonk:
You should be able to do something like `-framework FLEX.framework` in the **Other Linker Flags** build setting. I haven't looked into specifics, hopefully that sets you on the right path.

PR #159, opened 2016-11-04
Add Carthage instructions for FLEX exclusion from release builds

defagos:
Hi!

I mostly use Carthage for my projects, and needed a way to properly exclude FLEX from release builds. I found one and added instructions to the readme file. This provides a more detailed answer to issue #131, which has been closed.

I did not include any screenshot since yours seem hosted on your own. Here is an example if you want to upload it with the others and update the readme file accordingly:

flex_exclusion

Thanks for considering this (small) contribution to your awesome project!

Best regards.

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,8 +124,14 @@ Alternatively, you can manually add the files in `Classes/` to your Xcode projec
 
 
 ## Excluding FLEX from Release (App Store) Builds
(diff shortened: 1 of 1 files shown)


ryanolsonk:
Great! This will be super helpful to those using Carthage. Thanks for the contribution!

I no longer have access to the location that has been hosting the media files. Could we simply link to the image in this PR?

PR #160, opened 2016-11-07
Add image for Carthage FLEX exclusion setup

defagos:
Hi!

As suggested in pull request #159 comments, I added an example image to the Carthage setup instructions. I directly added the image in a `REAMDE-images` directory within the repository, instead of linking the image attached to my original PR.

Best regards.


ryanolsonk:
👍  thanks!

PR #161, opened 2016-11-08
Add a convenient action button in FLEXImagePreviewViewController

unixzii:
Hi, I've just replaced the original **Copy** button with an **Action** button in `FLEXImagePreviewViewController`, it behaves just like the **Photos** app, providing actions like "Save to Camera Roll", "AirDrop" and so on. I guess it's pretty handy.

I totally love this project, and hope to make it more powerful.


ryanolsonk:
Great addition. Thanks for the contribution!!

Issue #119, opened 2016-04-08
Crash Enumerating Heap Objects


sozorogami:
Trying to open the Heap Objects inspector in a large project, I get a crash on line 69 of `FLEXHeapEnumerator` with `EXC_BAD_ACCESS`. This occurs consistently on iOS 8.4 and 9.3, on both device and sim.

I believe this was working before with the current version of FLEX, so the issue may be caused by a class added recently. By throwing in a log on line 71 of `FLEXLiveObjectsTableViewController`, I can confirm that it enumerates through about 150000 objects successfully before crashing. Of course, the exception occurs before printing the name of the class it's choking on.

I would love advice as to how I can track down the object causing the crash.

yniesp:
+1

dissimiral:
+1

JALsnipe:
The crash occurs on line 69 of FLEXHeapEnumerator.m.

`zoneCount` is 7 but the app segfaults when i = 4.  Setting `zoneCount` to 3 in the debugger stopped the crash, but I still haven't narrowed down the reason why.  Investigating...

ryanolsonk:
I think the core issue here is that enumerating through the heap while the heap itself is changing is inherently unsafe. I'd like to investigate if there's a way we can make this async safe and signal to stop all threads. I don't have a timeframe on investigating that though.

ryanolsonk:
Actually, there is an issue that was introduced here that may cause issues for memory allocated through something other than malloc. I should have a fix up shortly.

ryanolsonk:
Can you see if things work any better following 22e7edb?

JALsnipe:
Still crashing at `zone->introspect->enumerator(TASK_NULL, (__bridge void *)block, MALLOC_PTR_IN_USE_RANGE_TYPE, (vm_address_t)zone, reader, &range_callback);`.

Backtrace:

```
* thread #1: tid = 0x263ed, 0x0000000122b027e0, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=2, address=0x122b027e0)
    frame #0: 0x0000000122b027e0
  * frame #1: 0x000000010ab56eaa FLEX`+[FLEXHeapEnumerator enumerateLiveObjectsUsingBlock:](self=FLEXHeapEnumerator, _cmd="enumerateLiveObjectsUsingBlock:", block=0x00007fff55173248) + 250 at FLEXHeapEnumerator.m:75
[...]

PR #169, opened 2017-01-30
Fix #168 and #119

### LINKED ISSUE #168: Crash in -[FLEXObjectExplorerViewController valueForPropertyAtIndex:]
Relevant part of the crash log:

    Last Exception Backtrace:
    0  __exceptionPreprocess + 0x7c
    1  objc_exception_throw + 0x38
    2  -[NSInvocation getArgument:atIndex:] + 0x200
    3  +[FLEXRuntimeUtility performSelector:onObject:withArguments:error:] + 0xae8
    4  +[FLEXRuntimeUtility valueForProperty:onObject:] + 0x154
    5  -[FLEXObjectExplorerViewController valueForPropertyAtIndex:] + 0x104
    6  -[FLEXObjectExplorerViewController subtitleForRow:inExplorerSection:] + 0x120
    7  -[FLEXObjectExplorerViewController tableView:cellForRowAtIndexPath:] + 0x474

Relevant disassembly (`+[FLEXRuntimeUtility performSelector:onObject:withArguments:error:]` + `0xae8` = `0x6f500`):

    0006f4e4    adrp       x8, #0xcb000      ; @selector(errorWithDomain:code:userInfo:)
    0006f4e8    add        x8, x8, #0x880    ; @selector(getReturnValue:)
    0006f4ec    ldur       x9, [x29, #-0xa0]
    0006f4f0    ldr        x2, [sp, #0x100]
    0006f4f4    ldr        x1, x8
    0006f4f8    mov        x0, x9
    0006f4fc    bl         imp___stubs__objc_msgSend
    0006f500    adrp       x8, #0xc9000

Since `@selector(getReturnValue:)` is in `x1` before the `msgSend`, isn't it crashing in `-[NSInvocation getReturnValue:]`?

I'm not sure what caused it but I've submitted a PR that fixes it somehow.

NSExceptional:
Reorganizes some code in `-[FLEXRuntimeUtility performSelector:onObject:withArguments:error]` and fixes a crash. See #168.

### DIFF
--- a/Classes/GlobalStateExplorers/DatabaseBrowser/FLEXTableContentViewController.m
+++ b/Classes/GlobalStateExplorers/DatabaseBrowser/FLEXTableContentViewController.m
@@ -147,6 +147,11 @@
         if ([obj2 objectForKey:text] == [NSNull null]) {
             return NSOrderedDescending;
         }
+        
+        if (![[obj1 objectForKey:text] respondsToSelector:@selector(compare:)] && ![[obj2 objectForKey:text] respondsToSelector:@selector(compare:)]) {
+            return NSOrderedSame;
+        }
+        
         NSComparisonResult result =  [[obj1 objectForKey:text] compare:[obj2 objectForKey:text]];
         
         return result;


ThePantsThief:
I've done nothing but merge for a while, can I squash those somehow?

ryanolsonk:
It's tricky to read what the functional change is in 58404e2 since it's mixed in with style/formatting changes. I'd like to see that separated out to properly evaluate the change. I'd also suggest rebasing the branch onto the latest master to eliminate the merge commits.

ThePantsThief:
> can I squash those somehow?

Rebasing is what I was looking for, I'll do that. Will that require a new pull request? I'm not super familiar with how that works

> I'd like to see that separated out to properly evaluate the change

As in just separate commits?

ryanolsonk:
Actually would be awesome if you want to make the changes as separate PRs so they can be evaluated/merged separately.

You should be able to force push to the branch after rebasing to update the existing PR. Alternatively you could open a new one if that's easier.

Thanks!

ThePantsThief:
Alrighty, that should do it for the original bug-fix

As for the stylistic changes, how do I get those in a separate PR? Haven't committed them yet because they'll show up here won't they?

ryanolsonk:
I'm hesitant to expand the scope of the try/catch blocks since it may be hiding a bug that could be fixed directly. Do you have a project with a consistent repro for the crash? I'd like to understand this one better.

ThePantsThief:
I do not. It was crashing in a third party app when scrolling through the properties.

The only thing I can think of is that `NSGetSizeAndAlignment` is clobbering `bufferSize` before throwing an exception, which results in the code below being executed no matter what.

AFAIK, it's actually bad practice to write code like this:

```objc
BOOL succeeded = NO;
@try {
    SomeExceptionalFunction();
    succeeded = YES;
    // ...
} @catch (NSException *exception) { }

if (succeeded) {
   foo();
else {
    failed();
}
```

It should instead be structured like this, because it's easier to read and you can be certain `foo()` will only execute if no exception was thrown:

```objc
@try {
    SomeExceptionalFunction();
    // ...
    foo()
} @catch (NSException *exception) {
    failed();
}
```

NSExceptional:
Bump

NSExceptional:
I don't know how to separate these commits into separate PRs, but I think they should both be merged anyway 😅

Classes/Utility/FLEXHeapEnumerator.m
```diff
@@ -71,8 +70,33 @@ + (void)enumerateLiveObjectsUsingBlock:(flex_object_enumeration_block_t)block
...
+            malloc_introspection_t *introspection = zone->introspect;
+
+            if (!introspection) {
+                break;
+            }
+
+            void (*lock_zone)(malloc_zone_t *zone)   = introspection->force_lock;
+            void (*unlock_zone)(malloc_zone_t *zone) = introspection->force_unlock;
+
+            // Callback has to unlock the zone so we freely allocate memory inside the given block
```
ryanolsonk: This is a pretty interesting approach. Are there not issues from allocating within the zone during enumeration?

ryanolsonk:
I'm definitely interested in seeing the change to heap enumeration merged! The other change is less clear to me. Expanding the scope of a try catch block may be covering a bug that we can fix more directly as I mentioned before. I'd like to have a repro case to better understand the issue before merging something there.

Classes/Utility/FLEXHeapEnumerator.m
NSExceptional: It would seem not! I can't find an app in which it still crashes. I'll ask around on /r/jailbreakdevelopers and /r/jailbreak to see if anyone can find crashes.

NSExceptional:
How can I split the changes into separate PRs? do I need to just make new branches for each set of changes and cancel this PR?

Classes/Utility/FLEXHeapEnumerator.m
ryanolsonk: Awesome! I would love to move forward with this piece of the PR

ryanolsonk:
Yea, that would be great. Thanks!

NSExceptional:
They're up!

NSExceptional:
> The other change is less clear to me. Expanding the scope of a try catch block may be covering a bug that we can fix more directly as I mentioned before. I'd like to have a repro case to better understand the issue before merging something there.

For anyone reading and wondering about this, I did find out what the specific crash was over in #183. I was going through my old PRs and realized I never mentioned this here.

PR #176, opened 2017-03-09
Fix #175

### LINKED ISSUE #175: Crash in [FLEXTableContentViewController multiColumnTableView:didTapHeaderWithText:sortType:] when the column represents a type which does not implement compare:
Tapping on a header in FLEXTableContentViewController to sort causes a crash if the property represents a type which does not implement 

`- (NSComparisonResult)compare:(Type)type`

such as a property that represents a RLMAccessor. It throws an exception with the following format:

`-[RLMAccessor_<SOME_NUMBER>_<REALM_OBJECT_CLASS_NAME> compare:]: unrecognized selector sent to instance`

NSIRLConnection:
Adds a check for `respondsToSelector:@selector(compare:)` to prevent a crash if the property corresponding to a column represents something that doesn't respond to `compare:`, e.g. a RLMAccessor. See #175.


ryanolsonk:
Thanks for the fix!

PR #177, opened 2017-03-24
Should keep the same completion logic

CodeLife2012:


CodeLife2012:
Anyone review this?

ryanolsonk:
I spend a while looking at this one and I think there's a problem introduced by the logic (as well as with the change in #68 which was previously merged). Requests without completion handlers will not be seen correctly by flex, regardless or whether network observing is enabled or not. I will take a crack at a deeper fix here.

ryanolsonk:
With further investigation this is sufficient. The request recording ends up getting picked up in the swizzled delegate method. Thanks for the fix!

PR #178, opened 2017-04-02
Fix crash in FLEXRuntimeUtility

tikoyesayan:
In AAPLAppDelegate section app crashes while parsing selector `URLSession:downloadTask:didFinishDownloadingToURL: `

in iOS 10 number of arguments returned by `method_getNumberOfArguments()` is more than in `selectorComponents` array which causes crash on line.
`NSString *prettyComponent = [NSString stringWithFormat:@"%@:(%@) ", [selectorComponents objectAtIndex:argIndex - kFLEXNumberOfImplicitArgs], readableArgType];`

In iOS 9 `method_copyArgumentType()` returns NULL which causes crash on line
`NSString *readableArgType = [self readableTypeForEncoding:@(argType)];`

### DIFF
--- a/Classes/Utility/FLEXRuntimeUtility.m
+++ b/Classes/Utility/FLEXRuntimeUtility.m
@@ -277,14 +277,22 @@ const unsigned int kFLEXNumberOfImplicitArgs = 2;
     NSMutableArray *components = [NSMutableArray array];
     
     NSString *selectorName = NSStringFromSelector(method_getName(method));
-    NSArray *selectorComponents = [selectorName componentsSeparatedByString:@":"];
-    unsigned int numberOfArguments = method_getNumberOfArguments(method);
+    NSMutableArray *selectorComponents = [[selectorName componentsSeparatedByString:@":"] mutableCopy];
     
-    for (unsigned int argIndex = kFLEXNumberOfImplicitArgs; argIndex < numberOfArguments; argIndex++) {
-        char *argType = method_copyArgumentType(method, argIndex);
-        NSString *readableArgType = [self readableTypeForEncoding:@(argType)];
+    // this is a workaround cause method_getNumberOfArguments() returns wrong number for some methods
+    if (selectorComponents.count == 1) {
+        return [selectorComponents copy];
+    }
+    
+    if ([selectorComponents.lastObject isEqualToString:@""]) {
+        [selectorComponents removeLastObject];
+    }
+    
+    for (unsigned int argIndex = 0; argIndex < selectorComponents.count; argIndex++) {
+        char *argType = method_copyArgumentType(method, argIndex + kFLEXNumberOfImplicitArgs);
+        NSString *readableArgType = (argType != NULL) ? [self readableTypeForEncoding:@(argType)] : nil;
         free(argType);
-        NSString *prettyComponent = [NSString stringWithFormat:@"%@:(%@) ", [selectorComponents objectAtIndex:argIndex - kFLEXNumberOfImplicitArgs], readableArgType];
+        NSString *prettyComponent = [NSString stringWithFormat:@"%@:(%@) ", [selectorComponents objectAtIndex:argIndex], readableArgType];
         [components addObject:prettyComponent];
     }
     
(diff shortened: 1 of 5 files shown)


Classes/Utility/FLEXRuntimeUtility.m
```diff
@@ -277,14 +277,16 @@ + (NSArray *)prettyArgumentComponentsForMethod:(Method)method
...
-    NSArray *selectorComponents = [selectorName componentsSeparatedByString:@":"];
-    unsigned int numberOfArguments = method_getNumberOfArguments(method);
+    NSMutableArray *selectorComponents = [[selectorName componentsSeparatedByString:@":"] mutableCopy];
+    if ([selectorComponents.lastObject isEqualToString:@""]) {
+        [selectorComponents removeLastObject];
+    }
     
-    for (unsigned int argIndex = kFLEXNumberOfImplicitArgs; argIndex < numberOfArguments; argIndex++) {
+    for (unsigned int argIndex = 0; argIndex < selectorComponents.count; argIndex++) {
         char *argType = method_copyArgumentType(method, argIndex);
```
piotr-tobolski: `argIndex` now starts from 0, not 2. Are you sure it is correct?

tikoyesayan: Since the `method_getNumberOfArguments()` returns not the correct number for some methods, namely `URLSession:downloadTask:didFinishDownloadingToURL`, I suggest that instead we get selector components and run the loop for this array. In this case there the crash connected with out of bounds of array will be eliminated. Also, as in the `for` loop you get the object at index `argIndex - kFLEXNumberOfImplicitArgs` for selector components, it is the same as to start from 0 until the count of selector components, as done in my version.

Classes/Utility/FLEXRuntimeUtility.m
```diff
@@ -277,14 +277,20 @@ + (NSArray *)prettyArgumentComponentsForMethod:(Method)method
...
     
     NSString *selectorName = NSStringFromSelector(method_getName(method));
-    NSArray *selectorComponents = [selectorName componentsSeparatedByString:@":"];
-    unsigned int numberOfArguments = method_getNumberOfArguments(method);
+    NSMutableArray *selectorComponents = [[selectorName componentsSeparatedByString:@":"] mutableCopy];
+    if ([selectorComponents.lastObject isEqualToString:@""]) {
+        [selectorComponents removeLastObject];
+    }
+    
+    if (selectorComponents.count == 1) {
```
ryanolsonk: Will this change the return value for single argument methods?

tikoyesayan:
I have changed the places of if statements in order to return right value for single argument methods.
I have also added a comment to explain why it should be done that way. 
Adding unit test is a good idea, now I am busy, when I am free I'll add unit tests.

tikoyesayan:
@ryanolsonk  I have added a unit test after my first commit. In the unit test I am checking if previous and new methods are returning the same value.

ryanolsonk:
Thanks for doing this and adding the test! Seems like the type information for that one method on `NSURLSessionDownloadDelegate` is pretty messed up in iOS 10.

Classes/Utility/FLEXRuntimeUtility.m
```diff
@@ -277,14 +277,22 @@ + (NSArray *)prettyArgumentComponentsForMethod:(Method)method
...
-    unsigned int numberOfArguments = method_getNumberOfArguments(method);
+    NSMutableArray *selectorComponents = [[selectorName componentsSeparatedByString:@":"] mutableCopy];
     
-    for (unsigned int argIndex = kFLEXNumberOfImplicitArgs; argIndex < numberOfArguments; argIndex++) {
-        char *argType = method_copyArgumentType(method, argIndex);
-        NSString *readableArgType = [self readableTypeForEncoding:@(argType)];
+    // this is a workaround cause method_getNumberOfArguments() returns wrong number for some methods
+    if (selectorComponents.count == 1) {
+        return [selectorComponents copy];
+    }
```
NSExceptional: This doesn't seem quite right. If a selector has one component, then it doesn't take any arguments. This change causes [every method screen to have a non-existent argument field.](https://i.imgur.com/W2CqMVw.png)

I think we should revert this in a new commit.

NSExceptional: @tikoyesayan @ryanolsonk

Classes/Utility/FLEXRuntimeUtility.m
NSExceptional: Yes, it does. His test did not accurately reflect this, as he adjusted the test to account for his own change.

PR #180, opened 2017-05-10
Add support for SQLCipher protected databases

revolter:
Related to #109.

### DIFF
--- a/Classes/GlobalStateExplorers/DatabaseBrowser/FLEXSQLiteDatabaseManager.m
+++ b/Classes/GlobalStateExplorers/DatabaseBrowser/FLEXSQLiteDatabaseManager.m
@@ -7,6 +7,7 @@
 //
 
 #import "FLEXSQLiteDatabaseManager.h"
+#import "FLEXManager.h"
 #import <sqlite3.h>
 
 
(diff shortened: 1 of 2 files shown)


Classes/GlobalStateExplorers/DatabaseBrowser/FLEXSQLiteDatabaseManager.m
```diff
@@ -33,6 +34,17 @@ - (BOOL)open {
         return YES;
     }
     int err = sqlite3_open([_databasePath UTF8String], &_db);
+
+#if SQLITE_HAS_CODEC
```
revolter: Preprocessor macro taken from [here](https://github.com/CocoaPods/Specs/blob/d0ec5a65e80656c8d78e12ff19f251df879e0bc2/Specs/1/2/7/SQLCipher/3.4.0/SQLCipher.podspec.json#L51).

revolter:
I don't know how I tested this, but it looks like the preprocessor macros are per pod, so the `SQLITE_HAS_CODEC` is only defined for the `SQLCipher` pod and is not visible by the `FLEX` pod.

You have to also add this hook:

	post_install do |installer|
		installer.pods_project.targets.each do |target|
			target.build_configurations.each do |config|
				if config.name == 'Debug'
					config.build_settings['GCC_PREPROCESSOR_DEFINITIONS'] ||= ['$(inherited)', 'SQLITE_HAS_CODEC=1']
				end
			end
		end
	end

PR #182, opened 2017-05-21
Fix #119

### LINKED ISSUE #119: Crash Enumerating Heap Objects

NSExceptional:


ryanolsonk:
Looks good, let's give it a shot!

PR #185, opened 2017-05-31
Add ability to filter out noisy network requests from being recorded

colinhumber:
Some network requests can be very noisy and congest the network recorder, such as analytics calls, or other calls you may not be interested in logged. This adds a simple filter to not log any of those requests.


ryanolsonk:
Nice! Thanks for the contribution!

PR #187, opened 2017-06-14
Fixed keyboard shortcuts in Xcode 9 simulator

nicked:
Pressing Cmd+S in the Simulator normally just takes a screenshot, but in Xcode 9, it also triggers the default "s" keyboard shortcut which enables the FLEX Select tool. This is because the keyboard flags is ANDed with a logical NOT of UIKeyModifierShift instead of the bitwise inverse


ryanolsonk:
Nice catch! Thanks for the fix!

PR #188, opened 2017-06-24
Simplify view controller presentation code

NSExceptional:
Removes duplicate code related to presenting a view controller from the
FLEX window (usually triggered by a toolbar button).

This adds `-[FLEXExplorerViewController
presentOrDismissViewControllerFromToolbar:shouldDismiss:completion:]`
which also makes it easier for others to add their own toolbar buttons
to present some other kind of screen, or for FLEX itself to utilize
again in the future.

### DIFF
--- a/Classes/ExplorerInterface/FLEXExplorerViewController.m
+++ b/Classes/ExplorerInterface/FLEXExplorerViewController.m
@@ -755,7 +755,7 @@ typedef NS_ENUM(NSUInteger, FLEXExplorerMode) {
 {
     // Note that we need to wait until the view controller is dismissed to calculated the frame of the outline view.
     // Otherwise the coordinate conversion doesn't give the correct result.
-    [self resignKeyAndDismissViewControllerAnimated:YES completion:^{
+    [self toggleViewsToolWithCompletion:^{
         // If the selected view is outside of the tap point array (selected from "Full Hierarchy"),
         // then clear out the tap point array and remove all the outline views.
         if (![self.viewsAtTapPoint containsObject:selectedView]) {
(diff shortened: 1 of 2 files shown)


NSExceptional:
Pardon all the whitespace changes by Xcode 😓

NSExceptional:
I agree, but I didn't want to completely toss some of the code that was already there without consulting you. Done!

Classes/ExplorerInterface/FLEXExplorerViewController.m
```diff
@@ -808,6 +808,19 @@ - (BOOL)wantsWindowToBecomeKey
...
 }
 
+- (void)toggleToolWithViewControllerProvider:(UIViewController *(^)())future completion:(void(^)())completion
+{
+    void (^presentBlock)() = ^{
+        [self makeKeyAndPresentViewController:future() animated:YES completion:completion];
+    };
+
+    if (self.presentedViewController) {
+        [self resignKeyAndDismissViewControllerAnimated:YES completion:presentBlock];
```
ryanolsonk: Ah, I missed something about this case. I was thinking we would unconditionally dismiss (and not present the new tool) in the case where something was already presented. As-is, I think there are some places where the intention will be to dismiss (e.g. in the `globalsViewControllerDidFinish:` callback) but we'll end up dismissing and re-presenting.

In thinking about this further, I'm not totally sure what the right thing to do is. Do you have any thoughts here?

NSExceptional: > I think there are some places where the intention will be to dismiss but we'll end up dismissing and re-presenting.

I think I designed it that way on purpose, but I can't remember why. Having just refreshed my memory with what's going on here, I think a sensible default would be to just dismiss the current tool if one is presented and do nothing else.

NSExceptional: (So, it no longer dismisses then immediately presents something if something is already presented)

NSExceptional:
Bump

Classes/ExplorerInterface/FLEXExplorerViewController.m
```diff
@@ -835,6 +835,19 @@ - (BOOL)wantsWindowToBecomeKey
     return self.previousKeyWindow != nil;
 }
 
+- (void)toggleToolWithViewControllerProvider:(UIViewController *(^)())future completion:(void(^)())completion
+{
+    void (^presentBlock)() = ^{
+        [self makeKeyAndPresentViewController:future() animated:YES completion:completion];
```
ryanolsonk: Slightly nicer to skip the block declaration and move this line into the `else` path

NSExceptional: Didn't even notice that the block wasn't being used in multiple places anymore. Definitely.

NSExceptional:
Fixed, ready to merge!

PR #189, opened 2017-06-24
Make FLEXExplorerToolbar more extensible

NSExceptional:
Exposes `FLEXExplorerToolbar.toolbarItems` as a public API so that others can more easily, safely, and reliably add their own items to the FLEX toolbar.

### DIFF
--- a/Classes/Toolbar/FLEXExplorerToolbar.h
+++ b/Classes/Toolbar/FLEXExplorerToolbar.h
@@ -12,6 +12,10 @@
 
 @interface FLEXExplorerToolbar : UIView
 
+/// The items to be displayed in the toolbar. Defaults to:
+/// globalsItem, hierarchyItem, selectItem, moveItem, closeItem
+@property (nonatomic, copy) NSArray<UIView*> *toolbarItems;
+
 /// Toolbar item for selecting views.
 /// Users of the toolbar can configure the enabled/selected state and event targets/actions.
 @property (nonatomic, strong, readonly) FLEXToolbarItem *selectItem;
--- a/Classes/Toolbar/FLEXExplorerToolbar.m
+++ b/Classes/Toolbar/FLEXExplorerToolbar.m
@@ -22,8 +22,6 @@
 
 @property (nonatomic, strong) UIImageView *dragHandleImageView;
 
-@property (nonatomic, strong) NSArray *toolbarItems;
-
 @property (nonatomic, strong) UIView *selectedViewDescriptionContainer;
 @property (nonatomic, strong) UIView *selectedViewColorIndicator;
 @property (nonatomic, strong) UILabel *selectedViewDescriptionLabel;
@@ -36,8 +34,6 @@
 {
     self = [super initWithFrame:frame];
     if (self) {
-        NSMutableArray *toolbarItems = [NSMutableArray array];
-        
         self.dragHandle = [[UIView alloc] init];
         self.dragHandle.backgroundColor = [FLEXToolbarItem defaultBackgroundColor];
         [self addSubview:self.dragHandle];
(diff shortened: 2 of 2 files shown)


Classes/Toolbar/FLEXExplorerToolbar.h
```diff
@@ -12,6 +12,10 @@
 
 @interface FLEXExplorerToolbar : UIView
 
+/// The items to be displayed in the toolbar. Defaults to:
+/// globalsItem, hierarchyItem, selectItem, moveItem, closeItem
+@property (nonatomic, copy) NSArray *toolbarItems;
```
ryanolsonk: Could you update the collection to be typed `<UIView>`?

Classes/Toolbar/FLEXExplorerToolbar.m
```diff
@@ -150,9 +138,31 @@ - (void)layoutSubviews
     descriptionLabelFrame.size.width = CGRectGetMaxX(self.selectedViewDescriptionContainer.bounds) - kHorizontalPadding - descriptionOriginX;
     self.selectedViewDescriptionLabel.frame = descriptionLabelFrame;
 }
+    
+    
+#pragma mark - Setter Overrides
 
+- (void)setToolbarItems:(NSArray *)toolbarItems {
```
ryanolsonk: Should we add a bail if `_toolbarItems == toolbarItems`?

Classes/Toolbar/FLEXExplorerToolbar.m
```diff
@@ -150,9 +138,31 @@ - (void)layoutSubviews
...
-#pragma mark - Setter Overrides
+    for (FLEXToolbarItem *item in toolbarItems) {
+        [self addSubview:item];
+    }
+
+    _toolbarItems = toolbarItems.copy;
+
+    // Lay out new items
+    [self setNeedsLayout];
+    [self layoutIfNeeded];
```
ryanolsonk: Let's remove this. It should be fine to let layout run without forcing it early.

NSExceptional:
Good points! I'll take care of those! As for the last one though, I think it makes sense for the UI to update as soon as you set the property. This is how all UIKit views behave (`UITabBar.items`, `UINavigationItem.rightBarButtonItems`, etc). There's probably never a time one wouldn't want the UI to update right away.

Imo the cost of updating the UI prematurely is negligible, but if you want I could just set the ivar directly in the initializer and add the subviews there too

Classes/Toolbar/FLEXExplorerToolbar.m
NSExceptional: I think it makes sense for the UI to update as soon as you set the property. This is how all UIKit views behave (UITabBar.items, UINavigationItem.rightBarButtonItems, etc). There's probably never a time one wouldn't want the UI to update right away.

Imo the cost of updating the UI prematurely is negligible, but if you want I could just set the ivar directly in the initializer and add the subviews there too

Classes/Toolbar/FLEXExplorerToolbar.m
NSExceptional: Done!

Classes/Toolbar/FLEXExplorerToolbar.h
NSExceptional: Done!

Classes/Toolbar/FLEXExplorerToolbar.m
ryanolsonk: I should clarify that the `setNeedsLayout` call is good and I only suggest removing the `layoutIfNeeded`. I poked into UITabBar out of curiosity, and didn't come across a `layoutIfNeeded` call in the surprisingly complicated path that comes from `setItems`.

ryanolsonk:
Gonna go ahead and merge this. We can consider removing the `layoutIfNeeded` call but it's also not doing any real harm. Thanks!

Classes/Toolbar/FLEXExplorerToolbar.m
NSExceptional: Ahhhhh gotcha, I see what you mean now. That could be removed, yeah

PR #190, opened 2017-06-29
Highlight request name in red if status code is a 4xx or 5xx

colinhumber:
Currently only if there's an error on the transaction will the name be displayed in red. Showing 4xx or 5xx errors in red is helpful when looking at a glance for failed requests.


ryanolsonk:
Awesome!

PR #191, opened 2017-07-12
NetworkHistory: Fix: NavigationBar disappeared. Keep filtered result …

cczallen:
…after clicked back from TransactionDetailPage.

In NetworkHistory page, searchBar is alway appeared, even if entered FLEXNetworkTransactionDetailTableViewController.

NavigationBar disappeared if entered detail page.
It can be back to history page by click "Cancel" first, but history filtered result has been cleared.

FLEXNetworkSettingsTableViewController: Change presenting  to push.

simulator screen shot 2017 7 12 4 57 17


ryanolsonk:
This appears to break if you do a search and then hit the back button. Upon returning to the network history list, search no longer works. Thanks for flagging the issue though! I'll take a crack at fixing this.

ryanolsonk:
Fixed in f7b4064

PR #198, opened 2017-10-08
Reveal exception reason in dialog

NSExceptional:
When a user-invoked method call fails, FLEX presents a dialog informing you that it failed, but does not say why.

In practice, it is more useful to see the exception name and reason than the (potentially crazy long) object description.


ryanolsonk:
:shipit:

PR #199, opened 2017-10-12
System log will cache messages

NSExceptional:
Currently, the system log screen just pulls in the entire list of messages from asl every time the list refreshes. Messages can sometimes expire quickly, which can be annoying when you're staring at one and it vanishes.

This new implementation will force log messages to persist until you dismiss the screen.


ryanolsonk:
👍

PR #204, opened 2017-11-19
Add generics to foundation collection classes

matrush:
Add generics to foundation collection classes so that the type is easier to reasoning about.


ryanolsonk:
Awesome! lgtm

PR #222, opened 2018-05-22
Fix heap enumeration crash

NSExceptional:
We only need to enumerate the `DefaultMallocZone` zone to find objects in the Objc runtime.

---

I have [several outstanding PRs.](https://github.com/Flipboard/FLEX/pulls/NSExceptional) Most of them are very simple. I don't know who else to contact about this except to just mention it here and hope someone important sees it. @matrush, could you take a look at some of them?


ryanolsonk:
Nice!

PR #223, opened 2018-05-28
Fix search bar

lavenntia:
fix search bar go off screen when focused on FLEXFileBrowserTableViewController.m & FLEXSystemLogTableViewController.m

.before.
screen shot 2018-05-28 at 11 37 45
.after.
screen shot 2018-05-28 at 13 34 45


ryanolsonk:
Do you have more detailed steps to reproduce this? I'm unable to reproduce (tried on iOS 11.4, both on device and in the simulator of an iPhone 7). Ideally we would not need to do this kind of manual intervention and UISearchController would handle things correctly itself. I'm open to a workaround if there's not a better alternative.

NSExceptional:
Closing for lack of activity and inability to reproduce the bug at this time. We can reopen this if @lavenntia responds with more information (iOS version, etc)

PR #226, opened 2018-07-06
add share  action to FLEXFileBrowserTableViewController

pujiaxin33:
Although FLEX can open files directly for browsing, such as databases, plists, text files, and so on. But for a large number of text information files, although FLEX provides copy function, a lot of text is difficult to transfer from my mobile phone to my computer. 
For example, I use 'CocoaLumberjack' to record the user log. When the tester feedbacks the problem, I need to export the log file to the mac. However, FLEX currently does not have file sharing capabilities. I think the file sharing feature is a must for some developers. So, I hope FLEX can support it.


ryanolsonk:
Cool!

PR #227, opened 2018-07-11
Group similar "objects with ivars referencing this object"

NSExceptional:
Often when looking through this area, it is time consuming to skip over objects you don't care about, such as auto-layout constraints when examining a UIView, among other things.

This makes it so that certain objects are grouped into their own sections. For now, Masonry and `NS*LayoutConstraint` objects are grouped together under **AutoLayout**, and `__NSObserver` and friends are grouped under **Trivial**. Other sections could easily be added.

[Here is what it looks like.](https://i.imgur.com/M19zkah.png)

@ryanolsonk Thoughts?

### DIFF
--- a/Classes/GlobalStateExplorers/FLEXInstancesTableViewController.m
+++ b/Classes/GlobalStateExplorers/FLEXInstancesTableViewController.m
@@ -12,17 +12,48 @@
 #import "FLEXRuntimeUtility.h"
 #import "FLEXUtility.h"
 #import "FLEXHeapEnumerator.h"
+#import "FLEXObjectRef.h"
 #import <malloc/malloc.h>
 
 
 @interface FLEXInstancesTableViewController ()
 
-@property (nonatomic, strong) NSArray *instances;
-@property (nonatomic, strong) NSArray<NSString *> *fieldNames;
+/// Array of [[section], [section], ...]
+/// where [section] is [["row title", instance], ["row title", instance], ...]
+@property (nonatomic) NSArray<FLEXObjectRef *> *instances;
+@property (nonatomic) NSArray<NSArray<FLEXObjectRef*>*> *sections;
+@property (nonatomic) NSArray<NSString *> *sectionTitles;
+@property (nonatomic) NSArray<NSPredicate *> *predicates;
+@property (nonatomic, readonly) NSInteger maxSections;
 
 @end
 
 @implementation FLEXInstancesTableViewController
+
+- (id)initWithReferences:(NSArray<FLEXObjectRef *> *)references {
+    return [self initWithReferences:references predicates:nil sectionTitles:nil];
+}
+
+- (id)initWithReferences:(NSArray<FLEXObjectRef *> *)references
+              predicates:(NSArray<NSPredicate *> *)predicates
+           sectionTitles:(NSArray<NSString *> *)sectionTitles {
+    NSParameterAssert(predicates.count == sectionTitles.count);
+
+    self = [super init];
+    if (self) {
+        self.instances = references;
+        self.predicates = predicates;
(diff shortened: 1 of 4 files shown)


Classes/GlobalStateExplorers/FLEXInstancesTableViewController.m
```diff
@@ -14,12 +14,17 @@
...
+typedef NSArray* Pair;
 
 @interface FLEXInstancesTableViewController ()
 
 @property (nonatomic, strong) NSArray *instances;
 @property (nonatomic, strong) NSArray<NSString *> *fieldNames;
 
+/// Array of [[section], [section], ...]
+/// where [section] is [["row title", instance], ["row title", instance], ...]
+@property (nonatomic, strong) NSArray<NSArray<Pair>*> *sections;
```
ryanolsonk: Might be nice to make the pair a strongly typed object with two fields

Classes/GlobalStateExplorers/FLEXInstancesTableViewController.m
```diff
@@ -74,21 +80,93 @@ + (instancetype)instancesTableViewControllerForInstancesReferencingObject:(id)ob
...
+               ([row hasPrefix:@"_NSAutoresizingMask"] && [row hasSuffix:@" _referenceItem"]) ||
+               [ignored containsObject:row];
+    };
+
+    BOOL(^isEssential)(Pair, NSDictionary *) = ^BOOL(Pair pair, NSDictionary *bindings) {
+        return !(isObserver(pair, bindings) || isConstraintRelated(pair, bindings));
+    };
+
+    switch (section) {
+        case 0: return [NSPredicate predicateWithBlock:isEssential];
```
ryanolsonk: Consider building a predicates array in `init` to remove the hardcoded `maxSections` count.

Classes/GlobalStateExplorers/FLEXInstancesTableViewController.m
```diff
@@ -74,21 +80,93 @@ + (instancetype)instancesTableViewControllerForInstancesReferencingObject:(id)ob
     FLEXInstancesTableViewController *instancesViewController = [[self alloc] init];
```
ryanolsonk: We can change the init to take arguments so that you can move row title construction inside there. I don't think we have a reason for these to be publicly settable.

Classes/GlobalStateExplorers/FLEXInstancesTableViewController.m
```diff
@@ -103,20 +181,27 @@ - (UITableViewCell *)tableView:(UITableView *)tableView cellForRowAtIndexPath:(N
...
+
     cell.textLabel.text = title;
     cell.detailTextLabel.text = [FLEXRuntimeUtility descriptionForIvarOrPropertyValue:instance];
     
     return cell;
 }
 
+- (NSString *)tableView:(UITableView *)tableView titleForHeaderInSection:(NSInteger)section {
+    switch (section) {
+        case 0: return nil;
```
ryanolsonk: Would be nice if this was a bit more driven by data we construct in init. Could have a section object with title + rows fields.

NSExceptional: I think my recent changes about wrap this up! Tested and working

NSExceptional:
Requested changes applied!

`FLEXInstanceDataSource` has a `.dataSourceClass` property that can be changed to a subclass of `FLEXInstanceDataSource` that overrides private methods to change it's behavior, such as if a developer wants more sections for their own types or something.

`FLEXInstancesTableViewController` has had much of it's clutter reduced as well.

PR #228, opened 2018-07-11
Additional object explorer scopes

NSExceptional:
Scopes now include:
- No inheritance (base class)
- Base class with parent attributes
- Base class with all inherited attributes, except NSObject
- NSObject attributes only

It is unusual that you need to see anything but the parent's attributes alongside those of the base class, and you especially rarely need to see NSObject attributes.

[Here's what it looks like.](https://i.imgur.com/sXMt7xT.png) The `+ Inherited` and `+ Parent` tabs hide depending on how far down the object is in the inheritance hierarchy (so, if it inherits directly from `NSObject`, the only other tab will be `NSObject`)

### DIFF
--- a/Classes/ObjectExplorers/FLEXObjectExplorerViewController.m
+++ b/Classes/ObjectExplorers/FLEXObjectExplorerViewController.m
@@ -17,6 +17,20 @@
 #import "FLEXInstancesTableViewController.h"
 #import <objc/runtime.h>
 
+typedef NS_ENUM(NSUInteger, FLEXObjectExplorerScope) {
+    FLEXObjectExplorerScopeNoInheritance,
+    FLEXObjectExplorerScopeWithParent,
+    FLEXObjectExplorerScopeAllButNSObject,
+    FLEXObjectExplorerScopeNSObjectOnly
+};
+
+typedef NS_ENUM(NSUInteger, FLEXMetadataKind) {
+    FLEXMetadataKindProperties,
+    FLEXMetadataKindIvars,
+    FLEXMetadataKindMethods,
+    FLEXMetadataKindClassMethods
+};
+
 // Convenience boxes to keep runtime properties, ivars, and methods in foundation collections.
 @interface FLEXPropertyBox : NSObject
 @property (nonatomic, assign) objc_property_t property;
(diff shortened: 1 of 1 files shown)


Classes/ObjectExplorers/FLEXObjectExplorerViewController.m
```diff
@@ -294,11 +376,14 @@ - (void)updateIvars
...
     self.ivars = [[self class] ivarsForClass:class];
-    self.inheritedIvars = [[self class] inheritedIvarsForClass:class];
+    self.ivarsWithParent = [self.ivars arrayByAddingObjectsFromArray:[[self class] ivarsForClass:[class superclass]]];
+    self.inheritedIvars = [self.ivars arrayByAddingObjectsFromArray:[[self class] inheritedIvarsForClass:class]];
+    self.NSObjectIvars = [[self class] ivarsForClass:[NSObject class]];
 }
 
 + (NSArray<FLEXIvarBox *> *)ivarsForClass:(Class)class
 {
+    if (!class) return @[];
```
ryanolsonk: nit: curly braces always

Classes/ObjectExplorers/FLEXObjectExplorerViewController.m
```diff
@@ -373,29 +456,36 @@ - (void)updateMethods
...
 - (void)updateFilteredClassMethods
 {
-    self.filteredClassMethods = [self filteredMethodsFromMethods:self.classMethods inheritedMethods:self.inheritedClassMethods areClassMethods:YES];
+    NSArray<FLEXMethodBox *> *candidateMethods = [self metadata:FLEXMetadataKindClassMethods forScope:self.scope];
+    self.filteredClassMethods = [self filteredMethodsFromMethods:candidateMethods areClassMethods:YES];
 }
 
 + (NSArray<FLEXMethodBox *> *)methodsForClass:(Class)class
 {
+    if (!class) return @[];
```
ryanolsonk: nit: curly braces always

Classes/ObjectExplorers/FLEXObjectExplorerViewController.m
```diff
@@ -36,25 +50,30 @@ @interface FLEXMethodBox : NSObject
 @implementation FLEXMethodBox
 @end
 
-static const NSInteger kFLEXObjectExplorerScopeNoInheritanceIndex = 0;
-static const NSInteger kFLEXObjectExplorerScopeIncludeInheritanceIndex = 1;
-
 @interface FLEXObjectExplorerViewController () <UISearchBarDelegate>
```
ryanolsonk: I think we could get the structure of the state on this class to a better place but it's not urgent. I'd also like to kill the inheritance used to customize the object explorers and do it via configuration. Would be cool to expose public hooks on the framework so you could customize display for different classes.

NSExceptional:
Ah yes, I forgot to fix the braces. I'm used to same-line braces.

Regarding object explorers, I think I commented about this on another issue but maybe not: I think objects should conform to some protocol that provides a custom explorer view controller if desired. That seems like a sleek solution to me

PR #258, opened 2019-04-08
Add address inspector

DrBreen:
* Add new item to FLEX menu - "Address Inspector"
* Allows inspecting/modifying objects at arbitrary memory addresses

### DIFF
--- a/Classes/GlobalStateExplorers/FLEXGlobalsTableViewController.m
+++ b/Classes/GlobalStateExplorers/FLEXGlobalsTableViewController.m
@@ -8,6 +8,7 @@
 
 #import "FLEXGlobalsTableViewController.h"
 #import "FLEXUtility.h"
+#import "FLEXRuntimeUtility.h"
 #import "FLEXLibrariesTableViewController.h"
 #import "FLEXClassesTableViewController.h"
 #import "FLEXObjectExplorerViewController.h"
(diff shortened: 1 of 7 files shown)


NSExceptional:
Could you post screenshots of the relevant screens, or a gif/video demonstration?

DrBreen:
@NSExceptional here's the demo:
https://streamable.com/tagbd

The UI is far from perfect, and I'd be glad to update it if you feel like it should be done better.

NSExceptional:
Nice! I have a suggestion though: when the row is tapped, instead of taking you to an entire screen with a search bar, a screen which will only ever have one result, I think we should instead present an alert view with a text field which takes an address, and have the associated warning as the alert message. Once the user pastes an address and hits ***Inspect***, the resulting object explorer is pushed.

I also think this should utilize the method I recently added (`-[FLEXRuntimeUtility pointerIsValidObjcObject:]`) to assert the validity of this pointer before pushing it. This could safely avoid crashes. Maybe the dialog should have three options though: Inspect, Unsafe Inspect, and Cancel. Unsafe Inspect could bypass the check.

Thoughts?

DrBreen:
I'm all in! This is a great suggestion! 
The only thing is - this method always fails on simulator due to `SUPPORT_NONPOINTER_ISA` being false. Is there a way to bypass it on simulator?

NSExceptional:
I didn't realize that, I'll look into it if you'll commit whatever code you were working towards with that. I should be able to fix it.

NSExceptional:
This is ready to merge, except for one caveat: sometimes, valid pointers are rejected by the `vm_read` call in `FLEXPointerIsReadable()`, even though `info.protection & VM_PROT_READ` is set.

For example, the `UIApplication` singleton. Try it. Meanwhile other objects seem to work fine, like the `FLEXManager` singleton. Using the `Unsafe Explore` option works for both.

NSExceptional:
Fixed! Merging.

PR #259, opened 2019-04-08
Show copy-ready address of inspected object

DrBreen:
* When inspecting object, show it's memory address even if description if not available
* Allow copying this address in the same manner as object description


DrBreen:
Pardon my debugging message in dealloc method.

NSExceptional:
Could you post a screenshot of what it looks like?

DrBreen:
Sure, of course!

DrBreen:
This is how it looks like in the object inspector. "Copy" tooltip appears on long press, just like with "Description" field

NSExceptional:
What do you think of this?

DrBreen:
That's a great idea! I love it, much cleaner solution.

PR #260, opened 2019-04-10
Fix safeArea for database content view

liaocm:
Currently, if you use iPhone X or any iPhone with a notch to view database, the column names will be hidden under the navigation bar. This is because the safeArea is not respected. This commit contains a simple fix to that.


Classes/GlobalStateExplorers/DatabaseBrowser/FLEXMultiColumnTableView.m
```diff
@@ -51,20 +51,24 @@ - (void)layoutSubviews
     CGFloat height = self.frame.size.height;
     CGFloat topheaderHeight = [self topHeaderHeight];
     CGFloat leftHeaderWidth = [self leftHeaderWidth];
+    CGFloat topInsets = 0.f;
     
+    if (@available (iOS 11.0, *)) {
```
matrush: let's add
```#if FLEX_AT_LEAST_IOS11_SDK```
check since FLEX is still supported on older Xcode & SDK.

liaocm:
Added compiler flag to support older SDKs

PR #263, opened 2019-04-11
Remove curl shortening

DrBreen:
* Copying curl request should not be trimmed, as long curl requests are prime target for copying - manually building them is tedious, and they should be available regardless of the size.


NSExceptional:
Curious, why did you open a new PR? 😅

DrBreen:
> Curious, why did you open a new PR? 😅

You know, I don't remember it, actually...
I remember there was a reason, but I can't really tell what was it 😐

PR #264, opened 2019-04-11
Remove copied curl shortening

DrBreen:
Copying curl request should not be trimmed, as long curl requests are prime target for copying - manually building them is tedious, and they should be available regardless of the size


NSExceptional:
I agree. Any limitations imposed are purely artificial up to some absurdly large request anyway

PR #265, opened 2019-04-12
podfile should include .mm file

lanbozhang:


NSExceptional:
I don't think we've actually updated the pod since I've added a `.mm` file, but I'll merge this anyway! Thanks!

PR #266, opened 2019-04-14
Add ability to use object-by-address for method arguments with type id

DrBreen:
Currently, if argument type is `id`, there's no way to pass anything that can't be parsed from JSON.
This change adds the ability to select the type of argument to pass:
* Value - works the same as before, JSON from input string is parsed to NSDictionary/NSArray/NSString/etc.
* Address - uses object at the specified address, so you can call any method with object you previously created and populated to your preference

Here's a simple demonstration of how it works:
https://streamable.com/hcwo6

### DIFF
--- a/Classes/Editing/ArgumentInputViews/FLEXArgumentInputObjectView.h
+++ b/Classes/Editing/ArgumentInputViews/FLEXArgumentInputObjectView.h
@@ -1,5 +1,5 @@
 //
-//  FLEXArgumentInputJSONObjectView.h
+//  FLEXArgumentInputObjectView.h
 //  Flipboard
 //
 //  Created by Ryan Olson on 6/15/14.
@@ -8,7 +8,6 @@
 
 #import "FLEXArgumentInputTextView.h"
 
-// #warning TODO This is never supported
-@interface FLEXArgumentInputJSONObjectView : FLEXArgumentInputTextView
+@interface FLEXArgumentInputObjectView : FLEXArgumentInputTextView
 
 @end
--- a/Classes/Editing/ArgumentInputViews/FLEXArgumentInputObjectView.m
+++ b/Classes/Editing/ArgumentInputViews/FLEXArgumentInputObjectView.m
@@ -1,15 +1,15 @@
 //
-//  FLEXArgumentInputJSONObjectView.m
+//  FLEXArgumentInputObjectView.m
 //  Flipboard
 //
 //  Created by Ryan Olson on 6/15/14.
 //  Copyright (c) 2014 Flipboard. All rights reserved.
 //
 
-#import "FLEXArgumentInputJSONObjectView.h"
+#import "FLEXArgumentInputObjectView.h"
 #import "FLEXRuntimeUtility.h"
 
-@implementation FLEXArgumentInputJSONObjectView
+@implementation FLEXArgumentInputObjectView
 
 - (instancetype)initWithArgumentTypeEncoding:(const char *)typeEncoding
 {
--- a/Classes/Editing/ArgumentInputViews/FLEXArgumentInputObjectView.m
+++ b/Classes/Editing/ArgumentInputViews/FLEXArgumentInputObjectView.m
@@ -1,5 +1,5 @@
 //
-//  FLEXArgumentInputObjectView.m
+//  FLEXArgumentInputJSONObjectView.m
 //  Flipboard
 //
 //  Created by Ryan Olson on 6/15/14.
@@ -8,6 +8,20 @@
 
 #import "FLEXArgumentInputObjectView.h"
 #import "FLEXRuntimeUtility.h"
+
+static const CGFloat kSegmentInputMargin = 4;
+
+typedef NS_ENUM(NSUInteger, FLEXArgInputObjectType) {
+    FLEXArgInputObjectTypeJSON,
+    FLEXArgInputObjectTypeAddress
+};
+
+@interface FLEXArgumentInputObjectView ()
+
+@property (nonatomic) UISegmentedControl *objectTypeSegmentControl;
+@property (nonatomic) FLEXArgInputObjectType inputType;
+
+@end
 
 @implementation FLEXArgumentInputObjectView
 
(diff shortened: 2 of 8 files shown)


NSExceptional:
You should be able to pass strings, arrays, dictionaries, or NSNumber to `id` currently:

```
"Strings must be in quotes"
["array", 1, 2, 3]
13457
```

As for addresses, I've wanted that myself but I can never remember why it's useful. Do you have a use case?

DrBreen:
Hm, I never considered that NSJSONSerialization is set up the way that it can parse fragmented JSON. Shouldn't have glossed over `NSJSONReadingAllowFragments` options. I will rework the input according to this new knowledge :)

Regarding addresses - sure, there is a whole lot of use cases. One such case may be, for example, UI tuning on-the-go - maybe the title doesn't look just right, so we want to pass new color and font to NSAttributedString. To do that, we should call `addAttribute:value:range:`, where value can be any object, basically. That's where ability to pass object by address comes in - we can alloc UIColor/UIFont/whatever we want, and pass it by address.

Other than that, where FLEX shines is in debugging weird bugs that are hard to reproduce - so if you caught something, and you don't have the ability to attach debugger to your device, you can use FLEX to inspect and play with bugged application state. By adding ability to pass any object to any method requiring `id` argument, we add another helpful tool to combat such bugs.

NSExceptional:
I don't believe we use NSJSONSerialization for plain strings or numbers but I could be wrong. I think we just check for quotes or something haha

> we can alloc UIColor/UIFont/whatever we want, and pass it by address

I see! Yeah, I've run into situations like that myself. Anyway, if you allocate something, copy it's address, then "go back" to use it somewhere, the object will have been deallocated 🙁 Maybe we need to give objects a "retain" button and add a "retained objects" section to the main menu? That way you can choose to manually keep it alive

Another alternative is if we had support for tabs we could just create an object and open a new tab without it deallocating. I'm down for either approach. Would you like to help with either?

DrBreen:
I'd be happy to help! This project is really interesting for me, and I'd be glad to contribute to it.

>I don't believe we use NSJSONSerialization for plain strings or numbers but I could be wrong. I think we just check for quotes or something haha
I just checked, and it's actually NSJSONSerialization that does all the heavy lifting with Strings and Numbers! You can see it in `FLEXRuntimeUtility.m`, in `objectValueFromEditableJSONString:` method - it's `NSJSONReadingAllowFragments` that makes the magic happen. 

But regarding this:
>Anyway, if you allocate something, copy it's address, then "go back" to use it somewhere, the object will have been deallocated 🙁
I also thought that's what happens, but this actually doesn't happen! You can reproduce it with current build with Address Inspector (see my other pull request) - try alloc/init-ing new object, or creating one with new, and copy it's address - you will see it's still alive. In Xcode this object is marked as "leaked", so it may be a bug, or it may be by design. So currently, passing address is be a cheap solution for "calling the method with any object" problem.

>Another alternative is if we had support for tabs we could just create an object and open a new tab without it deallocating.
However, this approach is a much cleaner alternative, and, if I understand correctly, it will allow allocating/setting up new instance of the object right where we call the method. Maybe we can add new tab for that, or button to allocate new object and paste address into input field.

DrBreen:
I found what causes the objects to live even after we leave the result screen of `[[ClassName alloc] init]`:
```
__unsafe_unretained id objectReturnedFromMethod = nil;
[invocation getReturnValue:&objectReturnedFromMethod];
returnObject = objectReturnedFromMethod;
```

If we remove `__unsafe_unretained`, then object is deallocated, but all kinds of weird crashes start happening.

NSExceptional:
That is interesting... you've uncovered a bug!

Can you do me a favor? 1) hold onto a crash log if you can't replicate the crash with the code the following code, 2) see if you can replicate the crash with this code:

```objc
void *objectReturnedFromMethod = nil;
[invocation getReturnValue:&objectReturnedFromMethod];
returnObject = (__bridge_transfer id)objectReturnedFromMethod;
```

DrBreen:
Sure, I'll check it out!
If we'll go with fixing the leak bug, then "Retain" button + list of retained object looks better, but tabs also look interesting. Maybe you can scribble a little sketch of how the tabs should look like, since we may be thinking of two different things.

DrBreen:
Ok, with the code you provided the crash can be replicated in sample application by:

1. Just selecting AAPLAppDelegate from FLEX menu
2. Opening `Buttons` from sample app menu, selecting `Button` text via FLEX menu, selecting `View Controller For Ancestor`, and then selecting `view` property.

In the first case, the crash occurs because we send `-[CFRunLoopTimer release]` to deallocated instance.
In the second case, the crash occurs because we send `-[UIButton respondsToSelector:]` to deallocated instance. 
These crashes are reproduced 100% of the time.

DrBreen:
Ok, the reason for the crash is that NSTimer from AAPLAppDelegate gets deallocated, my guess is due to `__bridge_transfer` - it takes on the object as it's own, then releases it, and when we actually try to access the object, it crashes.
 Simple `__bridge` does not have the same effect, but it also doesn't deallocate something we allocate manually.

DrBreen:
My take on this is that there's no easy fix (and maybe there's no fix at all, as in this is by design) - if we're taking retain/release onto ourselves by using `__bridge_transfer` - we mess up retain/release by ARC for the properties of the objects we inspect. If we use `__bridge` or if we use `__unsafe_unretained` - then we don't mess up ARC retain/release cycle for properties of inspected objects, but we don't have the ability to automatically dealloc objects we manually alloc.

I think that the approach that's currently implemented in FLEX is the correct one, since ARC is a compile time mechanism, and I don't see a sensible way to apply it to runtime inspection that FLEX is. 

Maybe what we can do is add some screen for orphaned objects, if it is even possible.

NSExceptional:
There should absolutely be a correct way to implement it; these methods are returning retained values, and I believe we are over-retaining them by not using `__bridge_transfer`.

That said, it is curious why this is having an effect on other objects entirely… I'll do some digging! If you find anything else, do share! 😄

NSExceptional:
So, this is super weird: it only crashes on the X* family of devices. Does not crash or leak on the 6s, 7, or 8. I wonder why that is.

NSExceptional:
Here's some [trace logs](https://github.com/Flipboard/FLEX/files/3082301/Archive.zip), one demonstrates a "leak" of a simple `Foo` class, when using the existing implementation, and the other demonstrates a zombie when using anything that makes an early release, like `__bridge_transfer` or `CFBridgingRelease`

So far I've been unable to diagnose it or find the root cause, but I think we're over-retaining something somewhere totally outside the scope of where we call `getReturnValue:`. Maybe in `potentiallyUnwrapBoxedPointer:` or `-[FLEXMethodCallingViewController actionButtonPressed:]`

DrBreen:
To simplify the inspection of the problem, I've created the repo:
https://github.com/DrBreen/RetainResearch

I've added you as collaborator in case you'll want to make some edits, or if you'll find something that you need to edit.

From what I see even with this barebones code the problem still occurs:
1. We hold the object in ViewController property
2. We get it via NSInvocation
3. We cast it's address via `__bridge_transfer`
4. It gets deallocated, even though it's still being held by `ViewController`

I'm looking into how can I see ARC byte code with retain/release calls.

DrBreen:
Tracing the calls with Instruments definitely shows additional `release` call when using `__bridge_transfer`! And assembly also shows the same.

Looks like by casting pointer to `id` with `__bridge_transfer` it acknowledges that now current scope is the owner of object, and it should release the object when it exits the scope.
 It doesn't happen with `__bridge` because compiler does not assume we are now owning the object - we are merely getting it to play and then we should put it back in place, the original owner knows better what to do with it and when to retain/release it.

So, I don't know if we can fix it. It looks like a fundamental problem - we can't transfer ownership of 3rd party objects to FLEX, since we are absolutely unaware of their lifecycles. Maybe we should give user two options - call the method without ownership transfer and with it?

NSExceptional:
Wow, awesome! Thanks! I'll definitely be using that to help debug this when I have more time.

> So, I don't know if we can fix it. It looks like a fundamental problem - we can't transfer ownership of 3rd party objects to FLEX, since we are absolutely unaware of their lifecycles. Maybe we should give user two options - call the method without ownership transfer and with it?

Again, I'm not sure I agree. Both the leak (using `__bridge` or `__unsafe_unretained`, which tell ARC to do nothing for the time being) and the crash (using anything that triggers a release, like `__bridge_transfer` or `CFBridgingRelease`) are related to objects whose lifetimes we should absolutely be able to reason about, as far as I know. We're just calling methods and looking at the return value. Why one approach leaks and one doesn't but crashes tells me we have a bug somewhere. We just need to find out where that extra retain or missing release is coming from :)

DrBreen:
I'm starting to check all the versions from 1.0.0 (similar to `git bisect` approach, or maybe I'll use the command, actually!), but I think we may be talking about two different things, that's where the disagreement comes from. But I'll check anyway, as there were a lot of cases where I was 99% I'm right, and in the end I was wrong :)

Now, consider everything that's written below correct if and only if we have no way to force `retain` call when casting `const void *` to `id`. If we have a way to do that, we may be able to bypass the problem I'm mentioning below. This bit was written after I wrote the whole comment as I suddenly got an idea - I'll be trying it now, I'm absolutely not sure if it'll work, but we may have a chance.

The demo I did with barebones code confirmed my suspicion that the extra release comes exactly from `__bridge_transfer`. According to Apple docs, `__bridge_transfer` is the same as [CFBridgingRelease](https://developer.apple.com/documentation/foundation/1587932-cfbridgingrelease). 

In the documentation they provide code sample:
```
CFStringRef cfName = ABRecordCopyValue(person, kABPersonFirstNameProperty);
NSString *name = (NSString *)CFBridgingRelease(cfName);
```

and they say this:
> You use this function to cast a Core Foundation-style object as an Objective-C object and transfer ownership of the object to ARC such that you don’t have to release the object

So basically, it's as if somebody allocated the memory for us, initialised the object and then gave it to us and said "here, take care of it, it's yours now". ARC sees this and adds these extra `release` you're talking about.

Now, how does it play with leaked object bug and with weird crashes we get when trying to inspect objects? 
Well, it's pretty simple - using `CFBridgingRelease` or `__bridge_transfer` we basically say to compiler - "okay, the object at this address is now mine, please handle it's memory cycle". Then, at some point (probably when we've build text description of property for UITableView cell in object inspector), the ARC sees that the object exits the scope, and sends `release` message to it. However, the only `retain` the object got was during it's allocation in it's owner - the object we're inspecting.

So it goes like this:
1. Object is allocated (retain count is 0)
2. Object is retained by it's owner (retain count is 1)
3. We inspect the object (retain count is 1)
4. We try to build property description - we get object via `NSInvocation` (retain count is 1)
5. We get the address of object and take ownership via `__bridge_transfer` (retain count is 1)
6. We finish building description (retain count is 1)
7. Object reference we're using for description leaves scope, and ARC sends `release` (retain count is 0)
8. We're trying to access the object for any other reason (inside original owner, we scrolled UITableView and got `cellForRow` call, etc.) - crash, as object is deallocated, yet it's reference in original owner still points to the address.

Now, regarding if we know the scope of the object being overreleased or not.
Actually, we're both right - we both know and we don't!

If we're talking specifically regarding new object allocation - the scope is definitely known for us - it should be alive from the time we allocated it to the point where we left the inspector screen of this object. 

However, if we're inspecting the object that doesn't belong to us, like property of `AAPLAppDelegate`, then we have absolutely no way to know it's scope. The property may live for as long as the object is alive, it may be deallocated after getting `NSNotification`, it may be allocated/deallocated on timer...there are endless possibilities!

So yes, you are right that when we allocate new object manually, we can know it's lifespan.
However, another problem arises - how do we differ between objects that we're allocating, and the objects that aren't ours? This is the fundamental problem I was talking about - other than few cases like '[[ClassName alloc] init]` or `[ClassName new]`, there's no surefire way to know whether the object we got is newly allocated and we can take ownership, or if it's the object with lifecycle that we can't control.

I'll give you an example. Let's say we define that all the properties we inspect - we get them using `__bridge`, without taking ownership, as it's pretty safe to say that properties are never ours, and we always take ownership of objects returned by methods using `__bridge_transfer`.

However, what if instead of inspecting property, we call the method of the same name?
Like, instead of `object.hisProperty` we'll call `[object hisProperty]` using FLEX. That way we don't overrelease the object if we're inspecting the property, but we're overreleasing it using the method, and we still get the crash. We can try to mitigate that, by, for example, comparing the name of method with properties, and if we find a match, we use `__bridge` instead of `__bridge_transfer`. But again, it's flaky - if getter name is different, then we're out of luck.

And if, for example, method returns one of object ivars - that way we'll still overrelease it and cause crash. 

That's why I believe there's a fundamental problem, and that's not a simple bug - it's just that we have no way of knowing whether the object we got can be safely released by us.

DrBreen:
Well, the method I tried didn't work - I though moving the cast to a different file with ARC disabled for that file with manual call to `retain` will help, but it didn't - the leak still happens.

NSExceptional:
> only if we have no way to force retain call when casting const void * to id

We do! We can `(__bridge CFType)` cast the object and pass it to `CFRetain`, or we can just cast it with `__bridge_retained` instead of `__bridge_transfer`. 

In a nutshell, `__bridge` does nothing, `__bridge_retained` retains and `__bridge_transfer` (auto)releases. I often turn to [this stackoverflow answer](https://stackoverflow.com/questions/14854521/where-and-how-to-bridge) when I get confused about these keywords.

> So it goes like this:

> 1. Object is allocated (retain count is 0)
> 2. Object is retained by it's owner (retain count is 1)
> 3. We inspect the object (retain count is 1)
> ...

That's not quite right. Did you ever look at the trace logs I uploaded? The object is retained and released several times throughout its lifetime.

> However, what if instead of inspecting property, we call the method of the same name?
Like, instead of object.hisProperty we'll call [object hisProperty] using FLEX. That way we don't overrelease the object if we're inspecting the property, but we're overreleasing it using the method

I'm a little confused here... properties are just syntax sugar around methods. Are you familiar with that? 😅

DrBreen:
>That's not quite right. Did you ever look at the trace logs I uploaded? The object is retained and released several times throughout its lifetime, which is what makes this so complicated :\

Are we talking about trace logs from FLEX, or from the basic example? If we're talking about FLEX, then I'd suggest moving to basic example you forked, since that is easier to work with and the problem is still there. Once we fix it there (or decide there's no fix), we can move on to fixing it in FLEX.

>I'm a little confused here... properties are just syntax sugar around methods. Are you familiar with that?

Of course I know what. What I wanted to show is the following case, but after I reviewed runtime introspection for properties, I learned that we have no problem knowing what getter and setter property has, even if it has different name.

This is the case I wanted to show:
`@property (readonly, getter=methodForTheTestProperty) NSString *testProperty;`

And we have the method definition:
```
- (NSString *)methodForTheTestProperty {
...
}
```

NSExceptional:
These (Instruments) trace logs: https://github.com/Flipboard/FLEX/files/3082301/Archive.zip

> Like, instead of object.hisProperty we'll call [object hisProperty] using FLEX.

This is what I was referring to when I asked about properties—these are the same thing.

> I learned that we have no problem knowing what getter and setter property has, even if it has different name.

Ah, I see

NSExceptional:
I seem to have gotten so caught up in the leak you uncovered that I forgot about the feature this PR was originally for!

A few constructive notes:

- Prefer the `NS_ENUM` macro over plain 'ol `enum`s
- Prefer `switch`ing over `self.someEnum` instead of `if-else`ing over possible values—the compiler can help catch missing cases if new ones are added later
- Prefer `const T XYZVar = 5;` over `#define XYZ 5`—variables will show up in the debugger but macros won't
- Properties are atomic by default (a mistake I'm sure someone regrets to this day), so you should always annotate them as `@property (nonatomic)`

Overall, nice work! I made some changes, like making sure git detected the file renames as such and not as add/delete, allowing us to specify a placeholder for both entry fields, and defaulting to the "preferred" method of entry based on the type of the property or argument being modified.

I have this on my local copy of your `object_input` feature branch. I only have commit access to your `master` branch, however. Can I push it to your `master` branch?

DrBreen:
Wow, thanks! I myself also forgot about this change due to having massive changes in my life going on. What was the actual cause for the leak?

NSExceptional:
I never did find out, but there is an open issue for it. It's not a super high priority for me now that we've spent hours debugging it to no avail.

If you have a developer account, it may be worthwhile to use one of your Code-Level Support tickets to help find the bug. (I've already used one of mine and I want to use the other on the Heap Enumerator because those APIs are poorly documented and some objects don't show up)

PR #273, opened 2019-05-09
Fix missing system frameworks dependencies

euanchan:
for detail, see https://github.com/Flipboard/FLEX/issues/272


euanchan:
I think some info will be useful if you wanna reproduce the error.

enable `use_frameworks!` and disable `CLANG_ENABLE_MODULES` in build-setting.

```
  post_install do |installer_representation|
    installer_representation.pods_project.targets.each do |target|
      target.build_configurations.each do |config|
        # trun off Enable Modules
        config.build_settings['CLANG_ENABLE_MODULES'] = 'NO'
      end
    end
  end
```

PR #275, opened 2019-05-14
Improve the detection of images

jnavarrom:
Not all images have extension. For example in our application thumbnails and previews are saved in a file using as name base 64 handle without extension.


NSExceptional:
In several places the `image` is re-initialized after we already have it loaded in memory. Could you try to consolidate the use of `imageWithContentsOfFile:` so that we're not calling it multiple times?

jnavarrom:
Oh, sorry, of course. Let me know if now it is correct

NSExceptional:
Did a little clean-up to make the diff smaller, we're good to merge now

PR #281, opened 2019-05-22
Link to the issues list

dillonwi:
I added a small section with a link to the known issues/bugs list.


ryanolsonk:
Thanks for looking for ways to improve the project! Since issues are already featured prominently as the second tab in the github UI, I think that will be sufficient for folks looking to find them.

PR #286, opened 2019-06-13
It'll be more convenient to enable share directory while browsing files.

euanchan:
After FLEX 3.0, it's great that share single file is enabled while under FLEXFileBrowser.

For our case, it would be more convenient if it's able to share a directory.

---

For share directory, it's needed to use UIDocumentInteractionController instead of UIActivityViewController.

### DIFF
--- a/Classes/GlobalStateExplorers/FLEXFileBrowserTableViewController.m
+++ b/Classes/GlobalStateExplorers/FLEXFileBrowserTableViewController.m
@@ -265,15 +265,9 @@
 {
     UIMenuItem *renameMenuItem = [[UIMenuItem alloc] initWithTitle:@"Rename" action:@selector(fileBrowserRename:)];
     UIMenuItem *deleteMenuItem = [[UIMenuItem alloc] initWithTitle:@"Delete" action:@selector(fileBrowserDelete:)];
-    NSMutableArray *menus = [NSMutableArray arrayWithObjects:renameMenuItem, deleteMenuItem, nil];
+    UIMenuItem *shareMenuItem = [[UIMenuItem alloc] initWithTitle:@"Share" action:@selector(fileBrowserShare:)];
 
-    NSString *fullPath = [self filePathAtIndexPath:indexPath];
-    NSError *error = nil;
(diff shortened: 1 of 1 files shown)


NSExceptional:
Unrelated to the PR: you should always try to submit a single or a few clean commits when making a PR. Take a look at the commit history of your PR [as it was submitted](https://user-images.githubusercontent.com/8371943/59443594-d8fa5600-8dc1-11e9-8954-1345014f37bf.png) and [after rebasing your only commit](https://user-images.githubusercontent.com/8371943/59443634-ed3e5300-8dc1-11e9-9fcd-0132b301ac7d.png) on top of Flipboard/master. Looks a lot cleaner, right? Rebase is pretty powerful!

NSExceptional:
Anyway, I don't think we want to _always_ present the document browser thing. We just want that for directories, right? Could you update your code to conditionally use the document thing instead of the share sheet, based on whether or not the item is a directory?

euanchan:
Thanks for your review.
for the files, when we use activityviewcontroller, while the file is raw(without extension, or can’t be recognized), the airdrop action will send the full path instead of the content of file. And use documentinteraction will cover those condtions.

NSExceptional:
Ahhh, I see. I think instead we need a specific activity which would present the document thingy. That's what I've seen other apps do. I personally still want access to the standard share sheet most of the time, but I agree we need both.

What do you think? I'm open to other ideas.

—
Tanner Bennett

euanchan:
Cause it’s late night here, I think I need to try a demo with more detail tomorrow, I’ll give more information after that.

NSExceptional:
I have fixed it so that the share sheet shares the actual file and not the file path. Gonna close this. Thanks for your contribution!

I will also be adding a way to long press to copy the file path in a future commit, as well as some images in the app's folder for anyone to test the share sheet with. Since they're in the app's folder, I'm thinking about adding a _"Browse app bundle"_ option to the main menu so you can actually get there. For now, I'm going to add a way to browse a bundle's files, so you can just go to the Main Bundle and tap _Explore bundle directory_

PR #289, opened 2019-06-18
Initial Dark Mode work

bdotdub:
This is an initial attempt to implement dark mode for FLEX.

In these two commits, we:
1. Move to FLEX semantic colors
2. For iOS13+, return a dynamic color provider for those semantic colors.

This dynamic color provider API is intentionally scoped only to handle light and dark mode (and not handling contrast, level, weight, etc) to keep things really simple.

This gets a bunch of the spots but there are still a few that still need updating, namely the icons.

### DIFF
--- a/Classes/Utility/FLEXColor.h
+++ b/Classes/Utility/FLEXColor.h
@@ -13,8 +13,8 @@ NS_ASSUME_NONNULL_BEGIN
 
 @interface FLEXColor : NSObject
 
-+ (UIColor *)backgoundColor;
-+ (UIColor *)backgoundColorWithAlpha:(CGFloat)alpha;
++ (UIColor *)systemBackgroundColor;
++ (UIColor *)systemBackgroundColorWithAlpha:(CGFloat)alpha;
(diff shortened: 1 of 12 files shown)


Classes/Utility/FLEXColor.h
```diff
@@ -0,0 +1,26 @@
...
+NS_ASSUME_NONNULL_BEGIN
+
+@interface FLEXColor : NSObject
+
++ (UIColor *)backgoundColor;
++ (UIColor *)backgoundColorWithAlpha:(CGFloat)alpha;
++ (UIColor *)secondaryBackgroundColorWithAlpha:(CGFloat)alpha;
++ (UIColor *)scrollViewBackgroundColor;
+
++ (UIColor *)primaryTextColor;
```
ryanolsonk: `textColor` and `secondaryTextColor` might be nice to mirror the background color methods

bdotdub:
@ryanolsonk updated with feedback and fixed network request history as well

image

PR #296, opened 2019-07-25
Fix build errors when using UIKit for Mac (Project Catalyst)

revolter:
Basically replacing the deprecated UIWebView with WKWebView.


NSExceptional:
Nice work! Can you fix the indentation here and `git commit --amend` it into the same commit, then force push it (`-f`)?

https://github.com/Flipboard/FLEX/pull/296/files#diff-8f0d87319094996005322b101e917039R96

revolter:
Sorry 🤦‍♂ I checked the diff twice before pushing but I still missed it.

PR #304, opened 2019-08-12
Add support for new UIScene APIs

### LINKED ISSUE #301: FLEX doesn't work when using the new Scene APIs from iOS 13
I managed to get it working by using this:

```objc
FLEXManager *manager = [FLEXManager sharedManager];

UIWindow *window = [manager valueForKey:@"explorerWindow"];

if (!window.windowScene) {
	UIScene *scene = [UIApplication sharedApplication].connectedScenes.anyObject;

	if ([scene isKindOfClass:[UIWindowScene class]]) {
		window.windowScene = (UIWindowScene *)scene;
	}
}

[manager showExplorer];
```

so I think that there should simply be a public method that the client calls and pass his/her desired `UIWindowScene` to assign to FLEX's private `window`. What do you think?

revolter:
Fixes #301.

### DIFF
--- a/Classes/Manager/FLEXManager.m
+++ b/Classes/Manager/FLEXManager.m
@@ -7,6 +7,7 @@
 //
 
 #import "FLEXManager.h"
+#import "FLEXUtility.h"
 #import "FLEXExplorerViewController.h"
 #import "FLEXWindow.h"
 #import "FLEXGlobalsEntry.h"
@@ -77,6 +78,18 @@
 - (void)showExplorer
 {
     self.explorerWindow.hidden = NO;
+#if FLEX_AT_LEAST_IOS13_SDK
+    if (@available(iOS 13.0, *)) {
+        if (!self.explorerWindow.windowScene) {
+            for (UIScene *scene in [UIApplication sharedApplication].connectedScenes) {
+                if (scene.activationState == UISceneActivationStateForegroundActive) {
+                    self.explorerWindow.windowScene = (id)scene;
+                    break;
+                }
+            }
+        }
+    }
+#endif
 }
 
 - (void)hideExplorer
@@ -91,6 +104,16 @@
         [self hideExplorer];
     }
 }
+
+#if FLEX_AT_LEAST_IOS13_SDK
+- (void)showExplorerFromScene:(UIWindowScene *)scene
+{
+    if (@available(iOS 13.0, *)) {
+        self.explorerWindow.windowScene = scene;
+    }
+    [self showExplorer];
+}
+#endif
 
 - (BOOL)isHidden
 {
--- a/Classes/Manager/FLEXManager.m
+++ b/Classes/Manager/FLEXManager.m
@@ -86,7 +86,7 @@
                 // Look for an active UIWindowScene
                 if (scene.activationState == UISceneActivationStateForegroundActive &&
                     [scene isKindOfClass:[UIWindowScene class]]) {
-                    self.explorerWindow.windowScene = (id)scene;
+                    self.explorerWindow.windowScene = (UIWindowScene)scene;
                     break;
                 }
             }
--- a/Classes/Manager/FLEXManager.m
+++ b/Classes/Manager/FLEXManager.m
@@ -80,9 +80,12 @@
     self.explorerWindow.hidden = NO;
 #if FLEX_AT_LEAST_IOS13_SDK
     if (@available(iOS 13.0, *)) {
-        if (!self.explorerWindow.windowScene) {
+        // Only look for a new scene if the one we have isn't the active scene
+        if (self.explorerWindow.windowScene.activationState != UISceneActivationStateForegroundActive) {
             for (UIScene *scene in [UIApplication sharedApplication].connectedScenes) {
-                if (scene.activationState == UISceneActivationStateForegroundActive) {
+                // Look for an active UIWindowScene
+                if (scene.activationState == UISceneActivationStateForegroundActive &&
+                    [scene isKindOfClass:[UIWindowScene class]]) {
                     self.explorerWindow.windowScene = (id)scene;
                     break;
                 }
@@ -111,7 +114,7 @@
     if (@available(iOS 13.0, *)) {
         self.explorerWindow.windowScene = scene;
     }
-    [self showExplorer];
+    self.explorerWindow.hidden = NO;
 }
 #endif
 
--- a/Classes/FLEXManager.h
+++ b/Classes/FLEXManager.h
@@ -9,6 +9,10 @@
 #import <Foundation/Foundation.h>
 #import <UIKit/UIKit.h>
 
+#if !FLEX_AT_LEAST_IOS13_SDK
+@class UIWindowScene;
+#endif
+
 typedef UIViewController *(^FLEXCustomContentViewerFuture)(NSData *data);
 
 @interface FLEXManager : NSObject
@@ -20,6 +24,10 @@ typedef UIViewController *(^FLEXCustomContentViewerFuture)(NSData *data);
 - (void)showExplorer;
 - (void)hideExplorer;
 - (void)toggleExplorer;
+
+/// Use this to present the explorer in a specific scene when the one
+/// it chooses by default is not the one you wish to display it in.
+- (void)showExplorerFromScene:(UIWindowScene *)scene API_AVAILABLE(ios(13.0));
 
 #pragma mark - Network Debugging
 


NSExceptional:
I think we need more than that to resolve #301. Namely, FLEX needs to be able to present itself somewhere automatically when `showExplorer` is called.

Is there some concept of a "first" scene? Or is there a way to get any one scene from a `UIApplication`?

revolter:
Well, you can get a set of the scenes and get a random one, but it's not a good idea. An app can have multiple scenes and FLEX can't correctly decide which one to use, so the only correct way is to let the user decide. I also don't know if you can find out when the first scene was created, which happens after didFinishLaunchingWithOptions from what I noticed.

NSExceptional:
Gotcha. After reviewing the documentation, I think the best course of action then is this:

1. Provide a new method, `showExplorerFromScene:`. I don't think it's a good idea for `FLEXManager` to hold onto a scene because scenes can come and go.

2. When `showExplorer` is called, we should simply find the first connected scene with role `UIWindowSceneSessionRoleApplication` and display the toolbar there. If there are multiple connected scenes with this role, we should log a warning to tell the developer to use `showExplorerFromScene:`.

What do you think?

revolter:
I don't quite agree with the first one. Currently, I only have one Scene, and in a new `SceneDelegate` class I set the scene on FLEX's window with the method I added, and then in the rest of the app I simply call `toggleExplorer` (I have multiple places in the app from where I can open FLEX).

With your suggestion, I'd have to either get the scene from `UIApplication`'s `NSSet`, either store the created scene from `SceneDelegate` as a public property and the access it from all the places, which is not that convenient. Also, FLEX is not necessarily storing the scene, FLEX's window is. We could also set FLEX's window as a `readonly` public property for example.

revolter:
I agree with the second point.

NSExceptional:
I was saying we should do both, not one or the other haha. If you only have one scene, then nothing changes. If you have multiple scenes *and* FLEXManager can't find the right one, only then would you have to use `showExplorerFromScene:`.

If you want to display multiple scenes with this specific role and need to have FLEX appear in a specific one, it should probably be your scene delegate which does the tracking of the scene's lifecycle instead of FLEX, as that's it's job anyway.

After some more digging, `UIScene.activationState` is `UISceneActivationStateForegroundActive` whenever a window is active. `FLEXManager` can always find the active scene with this!

```objc
UIScene *activeScene = nil;
for (UIScene *scene in [UIApplication sharedApplication].connectedScenes) {
    if (scene.activationState == UISceneActivationStateForegroundActive) {
        activeScene = scene;
        break;
    }
}
```

NSExceptional:
I've updated your PR, can you test it for me and see if `showExplorer` works like I predicted it would?

revolter:
I know, and it works in the commit body too, you don't have to set it in the title.

I'll test it tonight, if it works, I do think that it's the best approach as it's transparent for the user.

revolter:
Also, you shouldn't have force pushed my commit, you should have pushed a separate commit so the history remains clear, who did what, etc. In case someone needed to know how we got to the final solution or something. And, for example, for me to know what changed from my initial solution in case I forget.

NSExceptional:
> Also, you shouldn't have force pushed my commit, you should have pushed a separate commit so the history remains clear

I like to avoid cluttering the commit history with "partial" or rejected changes if we can avoid it, like in the case of a PR. However, now that you mention wanting to see what I changed, I agree that's a good way to help me help you guys, so I will start avoiding squashing my changes until the PR is ready to be merged!

As for this:

> In case someone needed to know how we got to the final solution or something.

I feel these threads serve that purpose to a degree. The discussion around the APIs is going to be more valuable than seeing some code added then removed right away, imo

revolter:
> I like to avoid cluttering the commit history with "partial" or rejected changes if we can avoid it, like in the case of a PR.

I noticed this. But I think that you're not aware of the ["Squash and merge"](https://help.github.com/en/articles/merging-a-pull-request#merging-a-pull-request-on-github) option, which does exactly this without modifying a fork's branch, which is actually not yours at all, it's author's branch.

For example, at fastlane, we're doing exactly this, and notice that [the log](https://github.com/fastlane/fastlane/commits/master) only has 1 commit for each PR, while the actual PRs still have all the commits, even after the merge. More so, the other merge options ("Create a merge commit" and "Rebase and merge") are completely disabled:

image

> I will start avoiding squashing my changes until the PR is ready to be merged!

So, with what I said, it also means that you shouldn't squash the commits on the PR's branch, and then force push it, even when the PR is ready to be merged. But instead, simply squash the commits from the PR's branch when merging in your own repo, using GitHub's button or from the command line (somehow).

> I feel these threads serve that purpose to a degree. The discussion around the APIs is going to be more valuable than seeing some code added then removed right away, imo

It depends. And, why remove something that you have? It's better to have something which you can ignore, instead of not having something that you'd like to know.

I genuinely hope that I didn't come as arrogant or condescending or something, and that you understand that I simply wanted to help you, without offending you 😃

NSExceptional:
I haven't been using that option because I didn't know how it would title and describe the singular commit. After looking it up just now, it appears to use the issue title, which is usually what I name the commit manually, so I'll start using it! (Semi-related: do you know how to accept a PR from the command line?)

> And, why remove something that you have? It's better to have something which you can ignore, instead of not having something that you'd like to know.

Just a matter of personal preference imo 😅I personally don't see the how the benefit of including an approach we decided not to take in the history outweighs a cleaner commit history, at least for something as minor as this.

> I genuinely hope that I didn't come as arrogant or condescending or something, and that you understand that I simply wanted to help you, without offending you 😃

You're good, I didn't read it that way! We're all here with good intentions. I learned something new and you showed me a better way to interact with PRs where one singular commit is desired. I hope I don't sound like my mind is set in stone on everything, I always try to read any advice with an open mind.

revolter:
> do you know how to accept a PR from the command line?

I didn't do it myself before, but I have found [this page](https://help.github.com/en/articles/checking-out-pull-requests-locally) for you, or you could use the [`hub` command line tool](https://hub.github.com/hub-merge.1.html).

> I assume you're fine with me simply rebasing PR branches before hitting Squash and Merge then? Because Squash and Merge won't do the rebase first, will it?

Rebasing a PR branch still implies a force push on the fork's branch, so not quite. Also, you don't need to rebase, except when there are conflicts and you want to solve them before merging.

I don't quite understand why you would want a rebase. I mean, the squashed commit containing all the changes in all the commits from the PR is obviously added "on the top", even if the PR was created from an older commit.

> I personally don't see the how the benefit of including an approach we decided not to take in the history outweighs a cleaner commit history, at least for something as minor as this.

Because it helps you not to make the same mistake. Or it can help you find out what you tried before reaching the current solution. It may not completely apply to this particular example, but it doesn't hurt to get in the habit of having as many commits as possible.

Also, knowing that we have the squash option, "outweighs a cleaner commit history" doesn't apply anymore 😛

Countless times I regretted doing some work in a single commit instead of having, apparently useless, intermediary commits, just to find out why I changed a line from 1000.

> I hope I don't sound like my mind is set in stone on everything, I always try to read any advice with an open mind.

No way! And even if you were hard-headed, I can totally understand it, as I'm like this too, sometimes 😅

Classes/Manager/FLEXManager.m
```diff
@@ -111,7 +114,7 @@ - (void)showExplorerFromScene:(UIWindowScene *)scene
     if (@available(iOS 13.0, *)) {
         self.explorerWindow.windowScene = scene;
     }
-    [self showExplorer];
+    self.explorerWindow.hidden = NO;
```
revolter: I see the bug before this change now 😆 Good catch!

revolter:
I added a small commit, but other than that, LGTM.

NSExceptional:
> I mean, the squashed commit containing all the changes in all the commits from the PR is obviously added "on the top", even if the PR was created from an older commit.

Is it? Or is a merge commit created? (The descriptions for these buttons are not very technical)

revolter:
> I don't see the harm if the branch is solely a PR feature branch

Because it can mean more than that. Remember, it is firstly simply a branch on a fork, which the author decided to request you pull from.

For example, I might create a fork, then a branch, do some stuff, and then use it in my project like:

```ruby
pod 'PodName', :git => 'https://github.com/revolter/PodName.git', :branch => 'branch_name'
```

And then, when you change something on it, you might break my setup. Remember, I just requested you to pull my changes (pull request), not modify them (push).

> I feel like rebasing has a bad rep compared to "merging" for whatever reason, and it doesn't deserve it.

Well, it kinda does. For one, it makes things hard for other people working on the same branch, as they won't be able to pull from the branch anymore.

So, if it's your own branch and you're the only person that it's working on it, then it's totally fine.

Here are 2 interesting texts written by more knowledgable people than me:
- [The Perils of Rebasing](https://git-scm.com/book/en/v2/Git-Branching-Rebasing#_rebase_peril)
- [Rebase vs. Merge](https://git-scm.com/book/en/v2/Git-Branching-Rebasing#_rebase_vs_merge)

And as I said too, I think that these are the most important phrases from the linked book:
- > Do not rebase commits that exist outside your repository and people may have based work on them.
- > In general the way to get the best of both worlds is to rebase local changes you’ve made but haven’t shared yet before you push them in order to clean up your story, but never rebase anything you’ve pushed somewhere.

> Rebasing requires force pushing, but only because it moves commits around, which is intended

It is intended, but not always desired, and as I previously said, it has some side-effects that you should be aware of.

For example, another issue would be referencing one of the commits. So, I create a branch, add a commit with SHA `a1b2c3`, and then reference it, maybe in a Wiki, or another issue or PR. And then, you rebase the branch, and the same commit will have the SHA `d4e5f6`, while existing references will remain unchanged, and they will point to the old commit instead of returning 404.

> Is it? Or is a merge commit created?

I don't know exactly, but, why does it matter if it's a merge commit (meaning that git will somehow remember that it was actually merged from commit x) or a simple commit, knowing that in both cases, the log will look the same? As I said, doing squash and merge looks exactly how you want:

image

One linear history with no merging branches.

> If they are squashed and put "on top" of Flipboard/master, that sounds like a rebase, right?

No. Rebase means changing an existing commit, by re-doing it but with a different hash. Squash and merge doesn't squash the original branch in-place, but rather creates a completely separate commit which only contains all the commits from the PR's branch squashed into one, and then adds it at the top of the branch you're merging on.

> that's just a merge but with one commit + the merge commit

No, that's just a merge with one commit, and that's all. Like I showed you in the previous screenshot from fastlane's log.

NSExceptional:
> And then, when you change something on it, you might break my setup. Remember, I just requested you to pull my changes (pull request), not modify them (push).

Well, if your concern is a maintainer making any changes at all to your PR branch, then you shouldn't have checked the "give maintainers commit access" box when you forked FLEX or opened the PR, or you should make a note in your PR that you don't want any maintainers to push to your PR branch. Making a PR will usually require changes unless it's very small. You probably shouldn't be using a PR branch for anything besides the PR for this very reason. Make a new branch based off of it if you need to, or be aware that you might have to deal with changes down the line. (This has nothing to do with rebasing so far.)

> I don't know exactly, but, why does it matter if it's a merge commit (meaning that git will somehow remember that it was actually merged from commit x) or a simple commit, knowing that in both cases, the log will look the same?

It won't necessarily look the same—a merge commit [connects two branches of a graph](https://stackoverflow.com/questions/20222732/why-does-git-merge-sometimes-create-a-commit-sometimes-does-not), and is not linear. Rebasing is linear. And according to the help site, [Squash and Merge does not rebase.](https://help.github.com/en/articles/about-merge-methods-on-github) :(

> No. Rebase means

😅I made myself very familiar with how it works and all the pros and cons before I chose to start using it. Semantically, rebasing does indeed "move" commits, and that's what I meant when I said that. But under the hood, like you said, it really just "replays" the changes in those commits and makes brand new commits out of them, because commit identity is a function of its location in the tree, among some other things.

> No, that's just a merge with one commit, and that's all.

Technically, yes: you can achieve this if a fast forward occurs. Semantically, a merge without a "merge commit" is equivalent to a rebase: if you have two branches `A` and `B` and merge them and the history ends up linear, then a fast forward occurred, which means either the branch with the changes, say `B`, was rebased on `A`, or that `A` had no new changes and did not diverge from `B`. If `A` _does_ diverge from `B`, however, then a merge commit _will always_ be generated unless you rebase one of the branches on top of the other. The stack overflow link above covers this as well

I have chosen to go with the [rebase workflow](https://www.atlassian.com/git/articles/git-team-workflows-merge-or-rebase) for this repo, because I value a linear commit history because it avoids [this](https://user-images.githubusercontent.com/8371943/63102867-895f1480-bf41-11e9-9638-686b270ab43d.png) confusing mess which makes it harder for someone exploring the repo's history to see what happened when. This means that at some point, **any PR is going to need to be rebased,** one way or the other. 😬

As the Atlassian article mentions, I realize rebase vs merge is a very controversial topic; there's generally no swaying someone from one side to the other—assuming each person is well informed on each approach—so do know that I don't expect you to agree with my stance here. I also realize it's not the right move for every repo; this would be a nearly impossible approach to take in a repo with a team working on it.

NSExceptional:
If you'll rebase your PR, I'll squash and merge it!

revolter:
You don't need to reference the issue in the title:

Screenshot 2019-08-16 at 19 19 42

revolter:
> you shouldn't have checked the "give maintainers commit access" box

The thing is that it's automatically checked :)) But yeah, you're right.

> Making a PR will usually require changes unless it's very small

In my personal experience, even if one character was needed, the maintainer asked the author to do it, even if he could have done it himself.

> You probably shouldn't be using a PR branch for anything besides the PR for this very reason. Make a new branch based off of it if you need to, or be aware that you might have to deal with changes down the line. 

True that.

> It won't necessarily look the same—a merge commit connects two branches of a graph, and is not linear. Rebasing is linear.

What do you mean by not being linear? It looks linear in the screenshot I previously sent:

image

All of those are merges from PRs.

> And according to the help site, Squash and Merge does not rebase.

No, but it does a fast forward merge, which is [kinda similar to a rebase](https://git-scm.com/docs/git-merge#_fast_forward_merge):

> In this case, a new commit is not needed to store the combined history; instead, the HEAD (along with the index) is updated to point at the named commit, without creating an extra merge commit.

> If `A` _does_ diverge from `B`, however, then a merge commit _will always_ be generated unless you rebase one of the branches on top of the other.

I think that this doesn't happen on GitHub. Because, as I said, at fastlane, there is no merge commit (no 2 parallel lines in the log), and surely there are cases in which the branches diverge. So I think that GitHub somehow takes care of it, so it can always fast-forward merge, meaning that it doesn't add an extra merge commit. And we also don't rebase anything, we just click the "Squash and merge" button, even for old PRs, and the log remains linear.

So I say you just try it for some PRs, and if it works, it'll save you and PR authors from extra work. And only ask the author to rebase in order to resolve conflicts, because they restrict you from merging the PR.

PR #309, opened 2019-08-18
Use Podfile and remove FLEX.xcworkspace

matrush:
This PR fixed the build error from a new checkout.

A lot of path like
```
path = ../../../../../../System/Volumes/Data/Users/tanner/Repos/FLEX/Classes/Core/FLEXCarouselCell.h;
```
was committed in https://github.com/Flipboard/FLEX/commit/d3ae20bebe819cc38081cb9c49a4878bd6bd4ae2, probably by accident.

I was not able to build `FLEX.xcworkspace` locally and spent a lot of time trying here and there.

In this PR I removed the `FLEX.xcworkspace` completely since one can use `pod install` to generate the workspace easily.

After this patch, I was able to build using
```
pod install
open FLEX.xcworkspace
select UICatalog project and build
```


ryanolsonk:
What was the build issue after fixing the paths? I like the idea of not requiring cocoapods to build the example but I'm open to it if there's ongoing pain here.

NSExceptional:
I feel the same way, but at the same time I realize it's an uphill battle with Xcode to avoid accidentally committing absolute paths like that into project/workspace files, so I see why it might be good to get it out of source control if we can.

Anyway, I'm getting a whole lot of Missing *.xcconfig warnings, and even an error, when I do a fresh clone and try to run the example out of the box. The files do exist, but it looks like Cocoapods is inserting the wrong path when configuring these settings. It's looking in `.../Pods/Pods/...` instead of just `.../Pods/...`. I'm not sure what's causing it

matrush:
@ryanolsonk let me try and report back what the error it. I wasn't very sure what's the recommended way of building the Examples,  but it indeed cost me tons of hours on it 😭

ryanolsonk:
Things appear to work for me after fixing the paths. I did build the flex framework first -- it's possible the example is not set up correctly to trigger a build of FLEX.framework as part of its own build.

matrush:
Ah actually that just fixed the build issue, maybe I did something wrong last time. I will just fix that for now if you feel strongly.

ryanolsonk:
👍 seems good to get the path fixes in alone. They're in the .xcodeproj anyways, which I think we have to keep around for Carthage support.

matrush:
Closing this in favor of #311.

PR #310, opened 2019-08-18
Use FLEXTypeEncoding enum and kFLEXUtilityAttribute whenever possible

matrush:
As titled, we should consolidate the usage of these enums throughout the codebase.


ryanolsonk:
Seems reasonable to me!

PR #311, opened 2019-08-18
Remove absolute paths from FLEX.codeproj

matrush:
Remove absolute paths from `FLEX.codeproj`.

I did a first attempt in #309, please see the discussion there about cocoapads.

This PR just fix the build issue without changing anything else.


ryanolsonk:
💯

NSExceptional:
Wow, I wonder how the heck those got in there. Maybe it was an Xcode 11 bug, I was using Xcode 11 for some of that. Thanks @matrush! Sorry 😅

PR #317, opened 2019-08-18
Add named struct field parsing support to fix the struct iVar issues

matrush:
Right now the struct parsing logic in FLEX doesn't support named fields.
A simple example like:
```
typedef struct TestSize {
    double width;
    double height;
} TestSize;
```
would be shown like this in FLEX:

The type encoding string is `{TestSize="width"d"height"d}`. However FLEX doesn't support the named fields in the parsing logic.

This PR tried to support the basic parsing support to make sure we can render both field name and field value properly. This would greatly help debugging C structs in general.

After the patch, this is how looks like for the same structs:

Most of the new code has inline comments.
Here are some references I looked at when implementing this:
- https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ObjCRuntimeGuide/Articles/ocrtTypeEncodings.html#//apple_ref/doc/uid/TP40008048-CH100-SW1
- https://github.com/opensource-apple/objc4/blob/cd5e62a5597ea7a31dccef089317abb3a661c154/runtime/objc-layout.mm
- https://github.com/nygard/class-dump/blob/master/Source/CDTypeParser.m

Also this depends on #315, otherwise the we will get wrong values.

### DIFF
--- a/Classes/Utility/FLEXRuntimeUtility.m
+++ b/Classes/Utility/FLEXRuntimeUtility.m
@@ -88,6 +88,19 @@ const unsigned int kFLEXNumberOfImplicitArgs = 2;
     return returnedObjectOrNil;
 }
 
++ (NSUInteger)fieldNameOffsetForTypeEncoding:(const FLEXTypeEncoding *)typeEncoding
+{
+    NSUInteger beginIndex = 0;
+    while (typeEncoding[beginIndex] == FLEXTypeEncodingQuote) {
+        NSUInteger endIndex = beginIndex + 1;
+        while (typeEncoding[endIndex] != FLEXTypeEncodingQuote) {
+            ++endIndex;
+        }
+        beginIndex = endIndex + 1;
+    }
+    return beginIndex;
+}
+
 + (NSArray<Class> *)classHierarchyOfObject:(id)objectOrClass
 {
     NSMutableArray<Class> *superClasses = [NSMutableArray new];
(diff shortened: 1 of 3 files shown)


Classes/Utility/FLEXRuntimeUtility.m
```diff
@@ -676,21 +689,63 @@ + (NSString *)readableTypeForEncoding:(NSString *)encodingString
     // See https://github.com/nygard/class-dump/blob/master/Source/CDType.m
     // Warning: this method uses multiple middle returns and macros to cut down on boilerplate.
     // The use of macros here was inspired by https://www.mikeash.com/pyblog/friday-qa-2013-02-08-lets-build-key-value-coding.html
-    const char *encodingCString = encodingString.UTF8String;
+    const char *encodingCString = [encodingString UTF8String];
```
NSExceptional: Why the change from property syntax to method call syntax?

Classes/Utility/FLEXRuntimeUtility.m
```diff
@@ -676,21 +689,63 @@ + (NSString *)readableTypeForEncoding:(NSString *)encodingString
...
+        return [NSString stringWithFormat:@"%@ %@", recursiveType, finalFieldNamesString];
+    }
 
     // Objects
-    if (encodingCString[0] == FLEXTypeEncodingObjcObject) {
+    if (encodingCString[0] == '@') {
         NSString *class = [encodingString substringFromIndex:1];
         class = [class stringByReplacingOccurrencesOfString:@"\"" withString:@""];
-        if (class.length == 0 || (class.length == 1 && [class characterAtIndex:0] == FLEXTypeEncodingUnknown)) {
+        if ([class length] == 0 || ([class length] == 1 && [class characterAtIndex:0] == FLEXTypeEncodingUnknown)) {
```
NSExceptional: Same as first, why the regression?

Classes/Utility/FLEXRuntimeUtility.m
```diff
@@ -676,21 +689,63 @@ + (NSString *)readableTypeForEncoding:(NSString *)encodingString
...
+                [finalFieldNamesString appendString:fieldName];
+            }
+        }
+        NSString *const recursiveType = [self readableTypeForEncoding:[encodingString substringFromIndex:fieldNameOffset]];
+        return [NSString stringWithFormat:@"%@ %@", recursiveType, finalFieldNamesString];
+    }
 
     // Objects
-    if (encodingCString[0] == FLEXTypeEncodingObjcObject) {
+    if (encodingCString[0] == '@') {
```
NSExceptional: Shouldn't we prefer our `FLEXTypeEncoding` enum to the literal `char`?

matrush: Must be a rebase issue :)

Classes/Utility/FLEXRuntimeUtility.m
matrush: I forgot why, probably because I’m working on a forked repo first and copied over some changes. I will revert all the syntax changes.

Classes/Utility/FLEXRuntimeUtility.m
NSExceptional: That's what I thought, wasn't sure though haha

PR #319, opened 2019-08-19
Try unwrap boxed value before set to input value for iVar and property input view

matrush:
Fixed `FLEXIvarEditorViewController` and `FLEXPropertyEditorViewController`

For instance variable type `const char *`, there are two issues in FLEX right now:
1. The `FLEXIvarEditorViewController` uses `FLEXArgumentInputNotSupportedView` because it didn't pass in its current value. And `FLEXArgumentInputStringView` only supports `NSString` type or `NSString` values.
2. The value will be a `NSValue` pointing to the `char *` pointer. 
However when we set this `NSValue` to `FLEXArgumentInputStringView`, it's not able to render it wasn't unwrapped.

This PR fixed both issues. The similar logic can be found here: https://github.com/Flipboard/FLEX/blob/bff9f1dd898fd723cacad39cd8014d13758e29f6/Classes/ObjectExplorers/Controllers/FLEXObjectExplorerViewController.m#L380-L381

### DIFF
--- a/Classes/Editing/FLEXIvarEditorViewController.m
+++ b/Classes/Editing/FLEXIvarEditorViewController.m
@@ -36,10 +36,13 @@
     [super viewDidLoad];
     
     self.fieldEditorView.fieldDescription = [FLEXRuntimeUtility prettyNameForIvar:self.ivar];
-    
-    FLEXArgumentInputView *inputView = [FLEXArgumentInputViewFactory argumentInputViewForTypeEncoding:ivar_getTypeEncoding(self.ivar)];
+
+    const char *typeEncoding = ivar_getTypeEncoding(self.ivar);
+    id currentValue = [FLEXRuntimeUtility valueForIvar:self.ivar onObject:self.target];
+    currentValue = [FLEXRuntimeUtility potentiallyUnwrapBoxedPointer:currentValue type:typeEncoding];
+    FLEXArgumentInputView *inputView = [FLEXArgumentInputViewFactory argumentInputViewForTypeEncoding:typeEncoding currentValue:currentValue];
     inputView.backgroundColor = self.view.backgroundColor;
-    inputView.inputValue = [FLEXRuntimeUtility valueForIvar:self.ivar onObject:self.target];
+    inputView.inputValue = currentValue;
     inputView.delegate = self;
     self.fieldEditorView.argumentInputViews = @[inputView];
     
--- a/Classes/Editing/FLEXPropertyEditorViewController.m
+++ b/Classes/Editing/FLEXPropertyEditorViewController.m
@@ -36,16 +36,18 @@
     [super viewDidLoad];
     
     self.fieldEditorView.fieldDescription = [FLEXRuntimeUtility fullDescriptionForProperty:self.property];
-    id currentValue = [FLEXRuntimeUtility valueForProperty:self.property onObject:self.target];
-    self.setterButton.enabled = [[self class] canEditProperty:self.property onObject:self.target currentValue:currentValue];
-    
+
     const char *typeEncoding = [FLEXRuntimeUtility typeEncodingForProperty:self.property].UTF8String;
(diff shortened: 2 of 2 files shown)


Classes/Editing/FLEXIvarEditorViewController.m
```diff
@@ -36,10 +36,13 @@ - (void)viewDidLoad
     [super viewDidLoad];
     
     self.fieldEditorView.fieldDescription = [FLEXRuntimeUtility prettyNameForIvar:self.ivar];
-    
-    FLEXArgumentInputView *inputView = [FLEXArgumentInputViewFactory argumentInputViewForTypeEncoding:ivar_getTypeEncoding(self.ivar)];
+
+    const char *typeEncoding = ivar_getTypeEncoding(self.ivar);
+    id currentValue = [FLEXRuntimeUtility valueForIvar:self.ivar onObject:self.target];
+    currentValue = [FLEXRuntimeUtility potentiallyUnwrapBoxedPointer:currentValue type:typeEncoding];
```
NSExceptional: I'm not sure we should try to unwrap NSValue instances here. It makes sense for object explorers because it's a read-only operation and you usually don't care about the "box" when you're trying to inspect something, but in this case if you happen to edit an ivar holding an NSValue which boxes a pointer to another object, the editor will display the unboxed object, and hitting "set" will change the ivar to the unboxed object even if the user didn't make any changes.

Thoughts?

Classes/Editing/FLEXPropertyEditorViewController.m
```diff
@@ -36,16 +36,18 @@ - (void)viewDidLoad
...
     
     self.fieldEditorView.fieldDescription = [FLEXRuntimeUtility fullDescriptionForProperty:self.property];
-    id currentValue = [FLEXRuntimeUtility valueForProperty:self.property onObject:self.target];
-    self.setterButton.enabled = [[self class] canEditProperty:self.property onObject:self.target currentValue:currentValue];
-    
+
     const char *typeEncoding = [FLEXRuntimeUtility typeEncodingForProperty:self.property].UTF8String;
-    FLEXArgumentInputView *inputView = [FLEXArgumentInputViewFactory argumentInputViewForTypeEncoding:typeEncoding];
+    id currentValue = [FLEXRuntimeUtility valueForProperty:self.property onObject:self.target];
+    currentValue = [FLEXRuntimeUtility potentiallyUnwrapBoxedPointer:currentValue type:typeEncoding];
```
NSExceptional: Same here

Classes/Editing/FLEXIvarEditorViewController.m
matrush: Good call on the setter issue.
So the 2 problem here:
1: without unwrapping, you basically cannot view the object/value at all since no proper input view could be found.
2: as you said, the setter is probably broken. Maybe we need some extra step there.

NSExceptional: I think we should special-case `FLEXArgumentInputStringView` to accept an `NSValue *` boxing a `char *`, unless someone wants to do some heavy-lifting and make `FLEXArgumentInputView` support unboxing and re-boxing `NSValue` instances, which might be a neat feature down the line but imo it's a very narrow use case outside of `char *`

matrush: I will think about this again and give it another try.

NSExceptional:
I have fixed this on my 4.0 branch on my own fork, so I will close this for now. FLEX now uses the string view for both `SEL` and `char *` arguments as long as it is either given an `NSString` or an `NSValue` whose `objCType` is the `SEL` or `char *` type.

For `SEL` and `char *`, the input field's `inputValue` always returns an `NSValue` wrapping whatever it is, and our `performSelector:withArguments:` helper method happily accepts this as-is and it all works flawlessly :)

PR #321, opened 2019-08-19
Add CGVector and NSDirectionalEdgeInsets support

matrush:
Add `CGVector` and `NSDirectionalEdgeInsets` support.

These are new common types. So following other examples to add support for them.


NSExceptional:
Awesome, looks good!

Possible future TODO: Do we know if the type encodings of all these structs include the field names and types? Is there a way we could just parse the type encodings to generate those field names instead of hard-coding them?

PR #325, opened 2019-08-21
Fix #324, crash when viewing NSNumber

### LINKED ISSUE #324: NSNumber wrapped boolean is causing crashes when tapping on the carousel bar’s NSNumber
66EBB60C-441E-4882-830E-4CEA33338309

Tapping on the NSNumber is crashing the app.

NSExceptional:
Now, we check for return length before even trying to invoke a method. If the length is 0 we abort. The length will be 0 for any unsupported type encoding such as that of `NSDecimal` or anything with bitfields or inline arrays.


matrush:
Sorry I missed the PR, I think it should fix the issue as you verified :)

PR #330, opened 2019-09-19
Obfuscate private API usage

### LINKED ISSUE #269: Non-public API usage:
I got this mail from App Store Connect.
My app is educational, so FLEX is available in the app store version too :)

Dear Developer,

We identified one or more issues with a recent delivery for your app, "SwiftHub - Git Client". Please correct the following issues, then upload again.

Non-public API usage:

The app references non-public selectors in Frameworks/FLEX.framework/FLEX: _canPerformAction:forCell:sender:, _performAction:forCell:sender:, _tableView, allObjects:, schema

If method names in your source code match the private Apple APIs listed above, altering your method names will help prevent this app from being flagged in future submissions. In addition, note that one or more of the above APIs may be located in a static library that was included with your app. If so, they must be removed.

khoren93:
Fixed #269


NSExceptional:
Sorry, my lack of comments in that area doesn't make this such an easy task I suppose.

We can't just rename those methods, because by implementing a method with the same name as a private method, **we're overriding that private method**. You'd have to use the Objc runtime to add the methods after manually constructing the selector, like we're already doing in [FLEXWindow.m](https://github.com/Flipboard/FLEX/blob/bff9f1dd898fd723cacad39cd8014d13758e29f6/Classes/ExplorerInterface/FLEXWindow.m#L49-L57)

Here are the steps in a nutshell:

1. Implement the method we wish to override under a different name (which you've done already by renaming it)
2. Find a place to put some code that will **run only once** to re-add the method under the desired name—a **private selector**. You'll probably want to put this code in `+load` for the class implementing the method. (In the example above we put it in `+initialize`, and after seeing it I've realized we have a bug; see [this article](https://www.mikeash.com/pyblog/friday-qa-2009-05-22-objective-c-class-loading-and-initialization.html) for more on these two methods)
3. Construct the **private selector** from a broken-up string, like we are in the example I linked.
4. Grab a reference to the `IMP` you wish to use to override the private method
5. Call `class_addMethod` with the private selector from step 3 and the `IMP` from step 4

If you have any questions about the Objc runtime, don't hesitate to ask! I'm here to help!

khoren93:
Thanks for Your reply,
I will try to study this topic better :)

NSExceptional:
Closing this because I'm about to release 4.0 and so much has changed by now that you'll have to do a major rebase or just reset your branch to `master` when 4.0 releases

PR #331, opened 2019-09-20
Fix heap enumeration on arm64e devices

opa334:
This was previously discussed here: https://github.com/NSExceptional/FLEXing/issues/7

This solution seems to work 100% of the time, so I am assuming that locking and unlocking the zones is no longer needed on arm64e. It might be a good idea to confirm it however.

(Only applies for apps that have an arm64e slice, arm64 apps running on an arm64e device are not affected)

### DIFF
--- a/Classes/Utility/Categories/NSObject+FLEX_Reflection.h
+++ b/Classes/Utility/Categories/NSObject+FLEX_Reflection.h
@@ -34,6 +34,8 @@ NSArray<FLEXProperty *> *FLEXGetAllProperties(_Nullable Class cls);
 /// @param instance used to mark methods as instance methods or not.
 /// Not used to determine whether to get instance or class methods. 
 NSArray<FLEXMethod *> *FLEXGetAllMethods(_Nullable Class cls, BOOL instance);
+/// @param cls a class object to get all instance and class methods.
+NSArray<FLEXMethod *> *FLEXGetAllInstanceAndClassMethods(_Nullable Class cls);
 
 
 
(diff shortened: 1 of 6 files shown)


NSExceptional:
So, the issue, like you pointed out originally, is that PAC uses those high order bits, so the "max realistic address" check is probably failing. Instead of getting rid of the check entirely, we just need to make the check work properly knowing that PAC uses those high bits.

My first idea is to check out the PAC APIs and see if there's a function to mask out the PAC bits.

My second idea is to use [a function I forgot I added that checks if a memory address is valid](https://github.com/Flipboard/FLEX/blob/master/Classes/Utility/FLEXObjcInternal.mm#L103) and readable. This is the problem I was trying to solve in the first place, so this seems perfect for what we need.

See if you can modify the check to use that function instead of a `<` comparison. (You'll need to add it to `FLEXObjcInternal.h` as it isn't exposed publicly right now).

As for why we need the check, it's my understanding that without locking the heap there is a race condition while we enumerate it. We might be given an address that is freed on another thread while we're about to read it, which causes nondeterministic and somewhat rare crashes. So it's no surprise to me that it would _appear_ removing the check works 100% of the time. It probably just exposes you to the rare chance of crashing, though.

opa334:
Yes, such a function exists, I previously used it to mask out the PAC, but even then, as I pointed out in the comment, it caused the thread to freeze up when calling (un)lock_zone. I will try to use it in addition to that function and see if that improves it.

opa334:
There may have been some other problem in FLEX that got fixed at some point that caused unlock_zone and lock_zone to not work on arm64e. I'm not sure exactly where it was fixed, but the check for adresses was still failing. I have improved the check and can confirm that heap enumeration and calling (un)lock_zone now works fine on my iPhone XS inside arm64e processes with the latest commits I added.

NSExceptional:
I've fixed it all, thank you for this!

PR #335, opened 2019-10-05
Gql warning popup

dev-aman:


NSExceptional:
Why is this necessary / useful?

PR #337, opened 2019-10-11
Add context menu to file browser in iOS 13.

elfredpagan:
The current implementation is deprecated, which limits the functionality
of the file browser.

### DIFF
--- a/Classes/GlobalStateExplorers/FileBrowser/FLEXFileBrowserTableViewController.m
+++ b/Classes/GlobalStateExplorers/FileBrowser/FLEXFileBrowserTableViewController.m
@@ -305,6 +305,45 @@
     // Since our actions are outside of that protocol, we need to manually handle the action forwarding from the cells.
 }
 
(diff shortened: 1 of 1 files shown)


ryanolsonk:
Thanks @elfredpagan!

NSExceptional:
A few things I noticed:

- This should be using `@available(iOS 13.0, *)` instead of preprocessor statements now that iOS 13 is out for real (I need to go back and remove all of those left over from the betas anyway)
- This PR doesn't touch the existing implementation which uses `UIMenuController`, right? What happens with both of them implemented at the same time like this?

ryanolsonk:
Please keep the macros around for now as we still have builds using the iOS 12 SDK. I believe keeping the UIMenuController implementation is to keep things working on iOS 12 and below, but it no-ops in 13.

NSExceptional:
Alrighty 👍🏼

Ah, they're both just table view delegate methods, that makes sense.

PR #344, opened 2019-11-06
Fix stack memory being used in async block

gl-lei:
The address of the messageText variable  is in the stack, when use disaptch_async, it will released.


NSExceptional:
Thanks! Nice catch!

PR #345, opened 2019-11-06
Address "This block declaration is not a prototype" warning

jnavarrom:
1. This block declaration is not a prototype
2. @available does not guard availability here; use if (@available) instead

We added FLEX manually to our project, we have the option `strict prototypes` enabled (YES) and we don't have in `Other warnings flags` the flag `-Wno-unsupported-availability-guard`

### DIFF
--- a/Classes/ObjectExplorers/Views/FLEXTableView.m
+++ b/Classes/ObjectExplorers/Views/FLEXTableView.m
@@ -18,13 +18,9 @@
 
 - (CGFloat)_heightForHeaderInSection:(NSInteger)section {
     CGFloat height = [super _heightForHeaderInSection:section];
-
-#pragma clang diagnostic push
-#pragma clang diagnostic ignored "-Wunsupported-availability-guard"
     if (section == 0 && self.tableHeaderView && !@available(iOS 13.0, *)) {
         return height - self.tableHeaderView.frame.size.height + 8;
     }
-#pragma clang diagnostic pop
 
     return height;
 }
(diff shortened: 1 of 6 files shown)


Classes/ObjectExplorers/Views/FLEXTableView.m
```diff
@@ -18,8 +18,10 @@ @implementation FLEXTableView
 
 - (CGFloat)_heightForHeaderInSection:(NSInteger)section {
     CGFloat height = [super _heightForHeaderInSection:section];
-    if (section == 0 && self.tableHeaderView && !@available(iOS 13.0, *)) {
-        return height - self.tableHeaderView.frame.size.height + 8;
+    if (section == 0 && self.tableHeaderView) {
+        if (@available(iOS 13.0, *)) {} else {
+            return height - self.tableHeaderView.frame.size.height + 8;
+        }
```
NSExceptional: This is the sort of thing I wanted to avoid with `-Wno-unsupported-availability-guard`. Could you just add that flag to your project instead? I really don't want ugly workarounds like this just to avoid a (pointlessly pedantic) warning :/

jnavarrom:
```
#pragma clang diagnostic push
#pragma clang diagnostic ignored "-Wunsupported-availability-guard"

#pragma clang diagnostic pop
```

^ What do you think if I add this? 
Developers can keep the flags in their projects as they want and we don't need to modify the if condition.

NSExceptional: I would really like to avoid needing all that extra code just to have some if statements look as they should. Personally, I think the warning is bogus and everyone should use this flag. I realize this is a huge OSS repo lots of people use though, so I want your feedback: is it asking too much for the minority of users who manually install FLEX into a project to add a flag to avoid this warning?

jnavarrom: ok, I will add it. Thanks for your time

PR #354, opened 2019-11-19
fixup tiny pedantic compilation issues

Daij-Djan:
This commit fixes some tiny pedantic compilation issues:
- availability ‘macro’ misuse
- header search paths

### DIFF
--- a/Classes/GlobalStateExplorers/FileBrowser/FLEXFileBrowserTableViewController.h
+++ b/Classes/GlobalStateExplorers/FileBrowser/FLEXFileBrowserTableViewController.h
@@ -6,9 +6,8 @@
 //  Based on previous work by Evan Doll
 //
 
-#import "FLEXTableViewController.h"
-#import "FLEXGlobalsEntry.h"
-#import "FLEXFileBrowserSearchOperation.h"
+#import <FLEX/FLEXTableViewController.h>
+#import <FLEX/FLEXGlobalsEntry.h>
 
 @interface FLEXFileBrowserTableViewController : FLEXTableViewController <FLEXGlobalsEntry>
 
--- a/Classes/GlobalStateExplorers/FileBrowser/FLEXFileBrowserTableViewController.m
+++ b/Classes/GlobalStateExplorers/FileBrowser/FLEXFileBrowserTableViewController.m
@@ -7,6 +7,7 @@
 //
 
 #import "FLEXFileBrowserTableViewController.h"
+#import "FLEXFileBrowserSearchOperation.h"
 #import "FLEXUtility.h"
 #import "FLEXWebViewController.h"
 #import "FLEXImagePreviewViewController.h"
--- a/Classes/GlobalStateExplorers/Globals/FLEXGlobalsEntry.h
+++ b/Classes/GlobalStateExplorers/Globals/FLEXGlobalsEntry.h
@@ -7,7 +7,8 @@
 //
 
 #import <UIKit/UIKit.h>
-#import "FLEXTableViewSection.h"
+#import <FLEX/FLEXTableViewSection.h>
+
 @class FLEXGlobalsTableViewController;
 
 typedef NS_ENUM(NSUInteger, FLEXGlobalsRow) {
--- a/Classes/ObjectExplorers/Views/FLEXTableView.m
+++ b/Classes/ObjectExplorers/Views/FLEXTableView.m
@@ -18,10 +18,13 @@
 
 - (CGFloat)_heightForHeaderInSection:(NSInteger)section {
     CGFloat height = [super _heightForHeaderInSection:section];
-    if (section == 0 && self.tableHeaderView && !@available(iOS 13.0, *)) {
+    if(@available(iOS 13.0, *)) {
+      return height;
+    }
+      
+    if (section == 0 && self.tableHeaderView) {
         return height - self.tableHeaderView.frame.size.height + 8;
     }
-
     return height;
 }
 


NSExceptional:
How are you installing flex in your project? These things do not affect anyone else normally.

Daij-Djan:
1
for availability we have very pendantic error flags... XCode complains 'availability not guarding ios availability' cause it is actually not meant to be used as bool part of a compound if

2.
The include errors happen when you dont use xcode but instead a CLI buildsystem like buck.
Header search paths are much more strict outside stock xcode(build)

Trying to use "A.h" from outside A's parent module is generally not kosher though. That holds true for C(++) too:
Thats why there IS a distinction between " and < ... 
"" = local
<> = non-local

:)

I guess it worked because 99.9% of ppl use xcode and not super strict errors

NSExceptional:
> cause it is actually not meant to be used as bool part of a compound if

I actually want to bring this up on the mailing lists. I think that warning is ridiculous; it was probably added to stop someone from doing this:

```objc
@available(…) {
    // code
}
```

But in practice all it does is prevent you from using it outside of simple if statements, which is a problem. If you could add the appropriate flag to disable it on a per-file basis—I think we only use it like that twice or so—I would be greatful if you could show me how you added it to your build system so I can add it to the README 😊 I have no plans to remove the flag which disables this warning, at least not until someone on the LLVM mailing lists can convince me it isn't a bogus warning.

As for import statements, I have a hunch there must be a way to add the appropriate header search paths to your build configuration as well. Either every header in FLEX needs to be included with <> or none should, not this weird middle ground where only a handful of import statements have been modified here. Can you show me the relevant build system code that adds FLEX to your project? This is something else I'd like to add to the README for anyone not using Xcode ☺️

Daij-Djan:
I can look into the warning and wrap the code in pragma push and pop I guess.. maybe that works :)

The import paths are CLEARLY wrong and I dont see us changing the configuration to work around.
The problem is also ONLY with those 4 includes ppl added. The <> are used correctly in most cases

To put it simply: Everytime you use quotes in an import in a public header, this is 100% wrong:
Everytime you use dont use quotes for an import in an implementation/private header: it is ok but you should only do so for external headers

NSExceptional:
I won't currently accept a PR that wraps the code in `#prama` directives, but you're welcome to do that on your own if you don't want to try adding specific flags for those files if that's what you want to do :(

The import paths are not wrong; all of those files are a part of the same project, so they should all be free to include one another without any path prefixes like `FLEX/`, and they shouldn't need `<>` when including one another. We only have [two public header files](https://github.com/Flipboard/FLEX/blob/master/FLEX.podspec#L38), and they both use the proper import statements: [FLEX.h#L9](https://github.com/Flipboard/FLEX/blob/master/Classes/FLEX.h#L9)

If you'd post your `BUCK` file, I can take a look to see what it would take to fix these without making a PR!

Daij-Djan:
ok

NSExceptional:
Okay! I really would like to help if I can. I want anyone to be able to use this project without Xcode, and it would be super helpful to us if I could get it working for you somehow.

Daij-Djan:
my comments stand. but is it is mainly that we make them public headers so we can reuse them for our extensions of flex:

e.g.

    [[FLEXManager sharedManager] registerGlobalEntryWithName:@"\U0001F5C4  App Group Container File Browser"
                                   viewControllerFutureBlock:^UIViewController *{
                                     NSURL *const containerURL = [[NSFileManager defaultManager] containerURLForSecurityApplicationGroupIdentifier:MSGAppGroupIdentifier()];
                                     if (containerURL) {
                                       return [[FLEXFileBrowserTableViewController alloc] initWithPath:[containerURL relativePath]];
                                     } else {
                                       UIAlertController *alertController = [UIAlertController alertControllerWithTitle:@"Not Available"

and more :)

NSExceptional:
Why not submit a PR to add an app group container browser?

Daij-Djan:
:D that was just one example.. still that can be a point for you but counter question: why offer an extension api that noone should use? :) 

is fixing 5 includes worth our time? this is not changing anything for you and 9/10 headers are already using <FLEX/ still imports so im not creating inconsistencies but closing them! :)

NSExceptional:
> why offer an extension api that no one should use?

It's just not public yet. But it will be eventually! Be aware that since these APIs are not public, they are subject to change until they become public.

> this is not changing anything for you and 9/10 headers are already using <FLEX/ style imports so im not creating inconsistencies but closing them even! :)

Which other headers are using `<FLEX/*.h>` style imports? Or even `<>`? The only one I see is `FLEX.h`—as intended—and the one from you recently merged PR, which I intend to change back.

> is fixing 5 includes worth our time?

Well, I've just spent a couple hours trying to get buck to even *build*—it doesn't build out of the box anymore apparently—all to no avail, because I couldn't get around some weird error. I would appreciate it if you could give me the opportunity to help you here so I can make it easier for anyone else who might be using Buck to use FLEX.

I've managed used FLEX without Xcode without any problems. I want to help you do the same.

Daij-Djan:
ah I assumed it was public by FlexManager.h:66 (#pragma mark - Extensions)
there also def. headers doing <FLEX/ to do EXACLTLY what I wanna do :D

just search for it and see FLEXGlobalsEntry.h and FLEXFileBrowserTableViewController.h

NSExceptional:
Oh! My bad, I was thinking of a different method. I'm doing a lot of refactoring on a private branch.

You can use that, but something like browsing app containers sounds like it would make a good addition to FLEX itself!

NSExceptional:
> just search for it and see FLEXGlobalsEntry.h and FLEXFileBrowserTableViewController.h

These are the only results I get with <bracketed> FLEX imports:

screenshot 2019-11-22 at 5 39 55 PM

Daij-Djan:
I am willing to add AppGroupContainer to FLEX:

itd mean a property on FLEXManager:
@property(copy) NSArray<NSString *> *appGroups

if thats agreeable, im going to put up a PR

Daij-Djan:
the other import we need back. Flex2 or whatever version had it cause we rely on it right now

Daij-Djan:
:) hope that is a compromise? :)

NSExceptional:
> the other import we need back. Flex2 or whatever version had it cause we rely on it right now

I am confused, what are you referring to here? Has a public header been made private since then or something?

Daij-Djan:
yes. if we plopp in a new version it doesnt compile. existing version does. 
I only tried to keep status quo

PR #357, opened 2019-11-23
fixup @available compilation issue and make all headers use <FLEX/*.h> syntax

Daij-Djan:
This commit fixes one @available compilation issue and then makes all headers use <FLEX/*.h> syntax.

this PR redos https://github.com/Flipboard/FLEX/pull/354 and unifies other headers

### DIFF
--- a/Classes/Editing/ArgumentInputViews/FLEXArgumentInputColorView.h
+++ b/Classes/Editing/ArgumentInputViews/FLEXArgumentInputColorView.h
@@ -6,7 +6,7 @@
 //  Copyright (c) 2014 Flipboard. All rights reserved.
 //
 
-#import "FLEXArgumentInputView.h"
+#import <FLEX/FLEXArgumentInputView.h>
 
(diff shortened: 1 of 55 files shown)


Daij-Djan:
sorry I totally agree this is wrong for the headers but thats what I thought was wanted. It doesnt make it impossible to compile with new code being added though .... or shouldnt def.

anyway I agree this is wrong and Ill just modify the headers locally

NSExceptional:
No, it does indeed not compile out-of-the-box if you copy the `Classes/` folder to another project. This is because the import statement `<FLEX/Header.h>` doesn't make sense when you're not compiling it as a module by itself; there is no `FLEX/` folder or FLEX module.

I advise against modifying the code locally like that. You really should try to adjust your build configuration to search for the right include paths. If you need help, feel free to ask. Again, **modifying the headers is not the right thing to do here**. You'll have to reapply your modifications each time we push changes on our end.

Daij-Djan:
right and right :)

PR #359, opened 2019-12-06
Fix object description selection

saitjr:


NSExceptional:
Can I ask what the benefit of this change is? I quite like the current `UIMenuController` implementation, which appears a little faster. The iOS 13 `UIMenu` stuff feels really slow for this

saitjr:
Long press the description of object, there was a Copy&Copy Address menu shown before. I found `FLEXObjectExplorerViewController` not implement `tableView:contextMenuConfigurationForRowAtIndexPath:point`. Just copy these code from `FLEXFileBrowserTableViewController`

NSExceptional:
Oh, was it not showing up for you on iOS 13 before you added this?

saitjr:
Yep, missing `tableView:contextMenuConfigurationForRowAtIndexPath:point` in `FLEXObjectExplorerViewController `

NSExceptional:
228de10 fixes this.

Our custom UIMenu callouts were not not appearing because the `UITextEffectsWindow` has a lower level than the FLEX window by default on iOS 13. Until a text field is activated, it would sit below the FLEX window, and custom callouts (copy, copy address) would not appear.

saitjr:
`UITextEffectsWindow` not always lower than `FLEXWindow`, and `UITextEffectsWindow` not always index at `1`.

This fix not work in some cases
```objc
UIApplication.sharedApplication.windows[1].windowLevel = self.view.window.windowLevel + 1;
```

saitjr:
How about find `UITextEffectsWindow` before set `windowLevel` ?

```objc
NSArray<UIWindow *> *textWindows = [UIApplication.sharedApplication.windows filteredArrayUsingPredicate:[NSPredicate predicateWithBlock:^BOOL(id  _Nullable evaluatedObject, NSDictionary<NSString *,id> * _Nullable bindings) {
    if ([evaluatedObject isKindOfClass:NSClassFromString(@"UITextEffectsWindow")]) {
        return YES;
    }
    return NO;
}]];
if (textWindows.count >= 0) {
    textWindows.firstObject.windowLevel = self.view.window.windowLevel + 1;
}
```
```

**Comment by NSExceptional** [comment]
Mmm, good call. I was too hasty with that fix. Your solution is much better. I will modify it to this and commit it:

```objc
NSArray<UIWindow *> *textWindows = UIApplication.sharedApplication.windows;
for (UIWindow *window in textWindows) {
    if ([window isKindOfClass:NSClassFromString(@"UITextEffectsWindow")]) {
        if (window.windowLevel <= self.view.window.windowLevel) {
            window.windowLevel = self.view.window.windowLevel + 1;
        }
    }
}
```

NSExceptional:
↑ I didn't mean to commit that just yet, whoops... I can edit it if need be

saitjr:
First one would be fine I think. If there were more than one `UITextEffectsWindow`, no matter `break` or though wasn't great solution. Expect find the `UITextEffectsWindow` which created by `FLEX`.

PR #365, opened 2019-12-27
Fix for color circle padding

duraidabdul:
There is currently almost no padding between the view backgroundColor preview circle and the info button. To fix this I've shifted the color preview over to the left by half the kContentPadding.


NSExceptional:
I have added these changes (and more to my 4.0 branch over at [NSExceptional/FLEX](https://github.com/NSExceptional/FLEX/tree/4.0)! Thanks!

PR #374, opened 2020-01-24
Fix crashes when previewing NSData keychain items.

jklausa:
:wave: 

This improves upon https://github.com/Flipboard/FLEX/pull/360, by also letting you view details of keychain items that have either `service`/`account` or `password` set to `NSData` types, incorrectly. 

One frequent cause of this, is integrating the Firebase framework (c.f https://github.com/firebase/firebase-ios-sdk/issues/4742), which creates a keychain entry with a `_pfo` keychain entry, but using an `NSData` object, instead of a `NSString`. 

This lets you view such items: 
Simulator Screen Shot - iPhone 11 Pro Max - 2020-01-24 at 15 12 01

Feel free to reformat/rename anything in the PR — and thanks for all the hard work on FLEX before :)

### DIFF
--- a/Classes/GlobalStateExplorers/Keychain/FLEXKeychainTableViewController.m
+++ b/Classes/GlobalStateExplorers/Keychain/FLEXKeychainTableViewController.m
@@ -83,8 +83,9 @@
     }
 
     if ([object isKindOfClass:[NSData class]]) {
-        return [NSString stringWithFormat: @"[NSData] %@", [[NSString alloc] initWithData:object
-                                                                                 encoding:NSUTF8StringEncoding]];
+        return [NSString stringWithFormat:@"[NSData] %@",
+            [[NSString alloc] initWithData:object encoding:NSUTF8StringEncoding]
+        ];
     }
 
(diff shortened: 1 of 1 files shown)


Classes/GlobalStateExplorers/Keychain/FLEXKeychainTableViewController.m
```diff
@@ -77,6 +77,19 @@ - (void)deleteItem:(NSDictionary *)item
...
 }
 
+- (NSString *)stringFromObject:(id)object {
+    if ([object isKindOfClass:[NSString class]]) {
+        return object;
+    }
+
+    if ([object isKindOfClass:[NSData class]]) {
+        return [NSString stringWithFormat: @"[NSData] %@", [[NSString alloc] initWithData:object
+                                                                                 encoding:NSUTF8StringEncoding]];
```
NSExceptional:
```suggestion
        return [NSString stringWithFormat:@"[NSData] %@",
            [[NSString alloc] initWithData:object encoding:NSUTF8StringEncoding]
        ];
```

jklausa:
Awesome, thanks :)

PR #386, opened 2020-02-21
feat: added ability to be compiled under project with extensions

tjerwinchen:
`UIApplication.sharedApplication` cannot be compiled under project with extension targets, so I make a tiny fix on this.

### DIFF
--- a/Classes/ExplorerInterface/FLEXExplorerViewController.m
+++ b/Classes/ExplorerInterface/FLEXExplorerViewController.m
@@ -167,6 +167,8 @@ typedef NS_ENUM(NSUInteger, FLEXExplorerMode) {
 
 - (void)viewWillTransitionToSize:(CGSize)size withTransitionCoordinator:(id<UIViewControllerTransitionCoordinator>)coordinator
 {
+    [super viewWillTransitionToSize:size withTransitionCoordinator:coordinator];
+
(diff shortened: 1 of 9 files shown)


NSExceptional:
What's the compilation error? Is that API unavailable in extensions? 😯

tjerwinchen:
Yes, UIApplication.sharedApplication is unavailable in extensions. 😂

tjerwinchen:
I found that in iOS 13, when a `presentationController` like  `FLEXGlobalsTableViewController` and `FLEXHierarchyTableViewController` dismissing with a swipe, `resignKeyAndDismissViewControllerAnimated:BOOL:completion` cannot get called. This will cause a bug because `keywindow` is not set back.  I make a tiny fix on this.

NSExceptional:
I have fixed this in the 4.0 branch on my fork, take a look

NSExceptional:
(Haven't gotten around to merging this PR yet but I will)

NSExceptional:
With your permission, I would like to refactor this PR into a compile-time check using preprocessor directives. I have my code ready to push to your fork and merge.

Unfortunately, Xcode does not provide a preprocessor variable to detect whether the target is an extension, so we will have to instruct users to add a define for `FLEX_TARGET_IS_EXTENSION` in their project settings.

What are your thoughts?

PR #389, opened 2020-03-05
Update FLEXNetworkHistoryTableViewController.m to fix  *** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason:*** -[__NSCFString rangeOfString:options:range:locale:]: nil argument

sunbohong:


NSExceptional:
This is already fixed in the 4.0 branch, please use that if you can. Or if you can wait, it will be pushed to master sometime soon.

PR #394, opened 2020-03-23
Make several functions static to avoid warning

matrush:
Several functions are not marked as static and it will report missing prototypes warning and failed to build if `-Wmissing-prototypes` flag is on.

Also fixed an issue with `-Wmissing-braces` flag on.


NSExceptional:
Thanks! Q: what's with the second pair of braces for the `rebind_symbols` argument?

Also, can we get rid of all the whitespace-only changes?

matrush:
@NSExceptional  Removing empty spaces should be better in this case to keep the file more consistent, right? Do we have a guidance? I'm seeing it's mixed currently so think it shouldn't matter.

NSExceptional:
Sure, but I'd rather not mix in whitespace changes with the rest of this stuff so that the significant changes are easier to see.

It probably needs to be done for the whole project anyway, I can go do it in a separate commit later?

matrush:
Sounds good, I will update :)

matrush:
Updated :)

PR #399, opened 2020-03-26
SyntaxHighlighting

jacobcxdev:
Working: @property
Not working: Everything else!

Also, there's a crash (which currently has a workaround implemented) in FLEXMetadataSection.m — see the FIXME.


matrush:
@jacobcxdev Can you provide more context, and provide some screenshots of your change? It's been a while since you opened the PR but we don't have enough context to get this merged.

jacobcxdev:
> @jacobcxdev Can you provide more context, and provide some screenshots of your change? It's been a while since you opened the PR but we don't have enough context to get this merged.

Sorry for the slow response :P. The idea was initially to add syntax highlighting support to FLEX by converting all the `NSString`s used to describe Objective-C stuff to `NSAttributedString`s, but it caused a bunch of crashes so I took a break and then totally forgot about the PR because I lost my jailbreak. With that said, I do still think that this would be a great feature for FLEX, but perhaps the usage of `NSAttributedString`s complicates things a little too much due to their inflexibility when compared to standard `NSString`s; maybe it would be a better idea to use a custom data structure.

PR #411, opened 2020-04-10
Added global entry with action to public extensibility.

alexito4:
As discussed [here](https://github.com/Flipboard/FLEX/issues/396#issuecomment-603916942) I've made the global entry with action publicly accessible.

I've moved some of my custom code to use this and it works 👍


NSExceptional:
At some point it seems I added this myself without realizing there was a PR for it, I'm sorry! Please reopen this if it is not working as expected or something

PR #417, opened 2020-04-24
Patch 1

ph661:
podspec add the weak_framework setting of SceneKit


NSExceptional:
I don't understand when something should be a weak framework or not. Why is this under weak and not just `frameworks` like the rest?

ph661:
> I don't understand when something should be a weak framework or not. Why is this under weak and not just `frameworks` like the rest?

Year, frameworks is ok.  I thought “SceneKit” is base on iOS 11， The original is On iOS 8

PR #421, opened 2020-05-21
Prevent buffer overflow from non-null-terminated response body

lexrus:
I got this page while I just tapped some random JSON response in the example project.
Screen Shot 2020-05-21 at 7 06 40 PM

And sometimes the JSON was NOT prettified even it looks well-formatted from the very beginning.
Screen Shot 2020-05-21 at 8 51 01 PM

Once I dismiss this view controller and push it again, like refreshing, it could be prettified most of the time.

I found that `prettyJSONStringFromData` returns `nil` randomly even though the server was not changed at all. In this case, there's a buffer overflow caused by a non-null-terminated string in this function.

You may read the following questions:
https://stackoverflow.com/questions/39754353/heap-buffer-overflow-on-nsdata-converting
https://stackoverflow.com/questions/2467844/convert-utf-8-encoded-nsdata-to-nsstring
https://stackoverflow.com/questions/27935054/determine-if-utf-8-encoded-nsdata-contains-a-null-terminated-string
https://medium.com/@hanru.yeh/information-leaks-on-nsjsonserialization-datawithjsonobject-data-2f8df25a8827

### DIFF
--- a/Classes/Utility/FLEXUtility.m
+++ b/Classes/Utility/FLEXUtility.m
@@ -341,9 +341,22 @@
     
     id jsonObject = [NSJSONSerialization JSONObjectWithData:data options:0 error:NULL];
     if ([NSJSONSerialization isValidJSONObject:jsonObject]) {
-        prettyString = [NSString stringWithCString:[NSJSONSerialization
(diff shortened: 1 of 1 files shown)


lexrus:
Close this PR since it's fixed in #491

PR #425, opened 2020-06-06
use __typeof instead of typeof to compile for Cxx instead of GNUxx

haoict:
I have problem when compiling this, and I think others may also have.
> Use __typeof instead of typeof. This is because both __typeof and typeof are extensions to C, but typeof is only enabled in Clang when the language is a GNUXX variant of C, but not for CXX.

https://stackoverflow.com/questions/32145326/weak-typeofself-weakself-self-expected-at-end-of-declaration/32145709#32145709


NSExceptional:
Thanks! Can you make it `__typeof__` though? That is the more widely available keyword

haoict:
@NSExceptional 
Sure, just updated ;)

haoict:
Hi
Do you have a plan to merge it or I can discard it? :)

NSExceptional:
My bad!

PR #436, opened 2020-07-07
Remove retain cycles from selectionHandler in several files

matrush:
Remove retain cycles from `selectionHandler` in several files.

These self were retained by the blocks `selectionHandler`, so we should use `weakSelf` and `strongSelf` instead.


NSExceptional:
Thanks!

PR #439, opened 2020-07-09
Move custom additions to the top

matrush:
I find myself using the custom additions much more frequent than built actions. I believe this is true to a lot of others as well. So this PR tries to move them to the top so they are more accessible.


NSExceptional:
I've been meaning to do this! Thanks!

NSExceptional:
Weird, I don't see the merge button on the Mobile app… hmm

PR #440, opened 2020-07-09
Add proper specifier for some properties

matrush:
These are raised up as warnings in Xcode, so I think fixing them would be better.


NSExceptional:
Can you squash and merge next time? I'd rather not have a lot of "fix code" commits when it should all just be one or two commits 😅

matrush:
> Can you squash and merge next time? I'd rather not have a lot of "fix code" commits when it should all just be one or two commits 😅

Ooops, how come I forget about it... Sorry wasn't using git much lately. Will make sure to address it next time...

NSExceptional:
No worries!

Anyway all these PRs look sweet, I'm gonna take a look at them tonight!

PR #441, opened 2020-07-09
Introduce FLEXTableRowDataViewController to view table row data

matrush:
Introduce `FLEXTableRowDataViewController` to view table row data.

Currently since FLEX 4.0, if you tap on a specific field on a db cell, the whole row's data will be aggregated and we only show the result in an alert view. This is not ideal since often times one wants to copy a certain value from the cell in order to achieve something else.

So I introduced a new view controller to focus on a certain row to view the data in a better way.

Demos:

### DIFF
--- a/Classes/GlobalStateExplorers/DatabaseBrowser/FLEXTableRowDataViewController.m
+++ b/Classes/GlobalStateExplorers/DatabaseBrowser/FLEXTableRowDataViewController.m
@@ -1,0 +1,54 @@
+//
+//  FLEXTableRowDataViewController.m
+//  FLEX
+//
+//  Created by Chaoshuai Lu on 7/8/20.
+//
+
+#import "FLEXTableRowDataViewController.h"
+#import "FLEXMutableListSection.h"
+#import "FLEXAlert.h"
+
+@interface FLEXTableRowDataViewController ()
+@property (nonatomic) NSDictionary<NSString *, NSString *> *rowsByColumn;
+@end
+
+@implementation FLEXTableRowDataViewController
+
+#pragma mark - Initialization
+
++ (instancetype)rows:(NSDictionary<NSString *, id> *)rowData {
+    FLEXTableRowDataViewController *controller = [self new];
+    controller.rowsByColumn = rowData;
+    return controller;
+}
+
+#pragma mark - Overrides
+
+- (NSArray<FLEXTableViewSection *> *)makeSections {
+    NSDictionary<NSString *, NSString *> *rowsByColumn = self.rowsByColumn;
+    
+    FLEXMutableListSection<NSString *> *section = [FLEXMutableListSection list:self.rowsByColumn.allKeys
+        cellConfiguration:^(UITableViewCell *cell, NSString *column, NSInteger row) {
+            cell.accessoryType = UITableViewCellAccessoryDisclosureIndicator;
+            cell.textLabel.text = column;
+            cell.detailTextLabel.text = rowsByColumn[column].description;
+        } filterMatcher:^BOOL(NSString *filterText, NSString *column) {
+            return [column localizedCaseInsensitiveContainsString:filterText] ||
+                [rowsByColumn[column] localizedCaseInsensitiveContainsString:filterText];
+        }
+    ];
+    
+    section.selectionHandler = ^(UIViewController *host, NSString *column) {
+        UIPasteboard.generalPasteboard.string = rowsByColumn[column].description;
+        [FLEXAlert makeAlert:^(FLEXAlert *make) {
+            make.title(@"Column Copied to Clipboard");
+            make.message(rowsByColumn[column].description);
+            make.button(@"Dismiss").cancelStyle();
+        } showFrom:host];
+    };
+
+    return @[section];
+}
+
+@end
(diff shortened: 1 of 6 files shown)


NSExceptional:
Working to get this merged

Classes/Core/Controllers/FLEXTableViewController.h
```diff
@@ -129,9 +129,9 @@ extern CGFloat const kFLEXDebounceForExpensiveIO;
 /// Subclasses may override. You should not need to call this method directly.
 - (void)setupToolbarItems;
 
-@property (nonatomic, readonly) UIBarButtonItem *shareToolbarItem;
-@property (nonatomic, readonly) UIBarButtonItem *bookmarksToolbarItem;
-@property (nonatomic, readonly) UIBarButtonItem *openTabsToolbarItem;
+@property (nonatomic, readonly, strong) UIBarButtonItem *shareToolbarItem;
+@property (nonatomic, readonly, strong) UIBarButtonItem *bookmarksToolbarItem;
+@property (nonatomic, readonly, strong) UIBarButtonItem *openTabsToolbarItem;
```
NSExceptional, line 134: `strong` is the default, I will be removing these

Classes/GlobalStateExplorers/DatabaseBrowser/FLEXTableRowDataViewController.m
```diff
@@ -0,0 +1,73 @@
...
+                return [column localizedCaseInsensitiveContainsString:filterText] ||
+                    [strongSelf->_valuesByColumn[column] localizedCaseInsensitiveContainsString:filterText];
+            } else {
+                return NO;
+            }
+        }
+    ];
+
+    _section.selectionHandler = ^(UIViewController *host, NSString *column) {
+        __strong __typeof(self) strongSelf = weakSelf;
```
NSExceptional, line 58: We have `@weakify`/`@strongify` macros I will be using instead

NSExceptional:
One of your commits, `Remove retain cycles from selectionHandler in several files`, caught one retain cycle, but the other one is not a retain cycle because `self` is a class object

NSExceptional:
I noticed you found you to do a lot of weakify/strongify-ing of `self` when using the APIs I designed. I was able to adjust them to not capture `self` at all. If you find yourself doing that a lot, please let me know, because either I need to improve the API so you don't need to do that, or you're probably doing it the hard way, haha 😄

NSExceptional:
As an example, this method:

```objc
- (NSArray<FLEXTableViewSection *> *)makeSections {
    NSArray<NSString *> *columns = _valuesByColumn.allKeys;
    __weak __typeof(self) weakSelf = self;
    _section = [FLEXMutableListSection list:columns
        cellConfiguration:^(UITableViewCell *cell, NSString *column, NSInteger row) {
            __strong __typeof(self) strongSelf = weakSelf;
            if (strongSelf) {
                cell.accessoryType = UITableViewCellAccessoryDisclosureIndicator;
                cell.textLabel.text = column;
                cell.detailTextLabel.text = [strongSelf->_valuesByColumn[column] description];
            }
        } filterMatcher:^BOOL(NSString *filterText, NSString *column) {
            __strong __typeof(self) strongSelf = weakSelf;
            if (strongSelf) {
                return [column localizedCaseInsensitiveContainsString:filterText] ||
                    [strongSelf->_valuesByColumn[column] localizedCaseInsensitiveContainsString:filterText];
            } else {
                return NO;
            }
        }
    ];

    _section.selectionHandler = ^(UIViewController *host, NSString *column) {
        __strong __typeof(self) strongSelf = weakSelf;
        if (strongSelf) {
            NSString *string = [strongSelf->_valuesByColumn[column] description];
            UIPasteboard.generalPasteboard.string = string;
            [FLEXAlert makeAlert:^(FLEXAlert *make) {
                make.title(@"Copied");
                make.message(string);
                make.button(@"OK");
            } showFrom:host];
        }
    };

    return @[_section];
}
```

Can be simplified like so, by capturing the data itself inside the closures instead of `self`. Because the data is immutable, we don't need a reference to `self` at all:

> _(and nor do we need to actually store the section in an ivar, we just need to vend it)_

```objc
- (NSArray<FLEXTableViewSection *> *)makeSections {
    NSDictionary<NSString *, NSString *> *rowsByColumn = self.rowsByColumn;
    
    FLEXMutableListSection *section = [FLEXMutableListSection list:self.rowsByColumn.allKeys
        cellConfiguration:^(UITableViewCell *cell, NSString *column, NSInteger row) {
            cell.accessoryType = UITableViewCellAccessoryDisclosureIndicator;
            cell.textLabel.text = column;
            cell.detailTextLabel.text = rowsByColumn[column].description;
        } filterMatcher:^BOOL(NSString *filterText, NSString *column) {
            return [column localizedCaseInsensitiveContainsString:filterText] ||
                [rowsByColumn[column] localizedCaseInsensitiveContainsString:filterText];
        }
    ];

    section.selectionHandler = ^(UIViewController *host, NSString *column) {
        NSString *string = rowsByColumn[column].description;
        UIPasteboard.generalPasteboard.string = string;
        [FLEXAlert makeAlert:^(FLEXAlert *make) {
            make.title(@"Copied");
            make.message(string);
            make.button(@"OK");
        } showFrom:host];
    };

    return @[section];
}
```

NSExceptional:
Thanks!

PR #442, opened 2020-07-09
Show full text for DB browser column headers

matrush:
Show full text for DB browser column headers.

Currently the header has a fixed width of 120pts. This isn't ideal as some of the columns will have names that cannot fit, and user cannot expand its width like Excel or Google Spreadsheet.

So in this PR I'm trying to use `MAX(120pt, sizeThatFits)` as the width of the header and content, so that they can be consumed in a better way.

### DIFF
--- a/Classes/GlobalStateExplorers/DatabaseBrowser/FLEXDBQueryRowCell.h
+++ b/Classes/GlobalStateExplorers/DatabaseBrowser/FLEXDBQueryRowCell.h
@@ -8,12 +8,21 @@
 
 #import <UIKit/UIKit.h>
 
+@class FLEXDBQueryRowCell;
+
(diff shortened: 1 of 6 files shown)


NSExceptional:
I have forced-pushed your branch to rebase your changes onto `master`.

To be honest, there is a lot going on here and I'm having trouble sifting through it all in so many separate commits. Since it seems like you backtracked at some point, I'd like to squash your commits into a single commit (with your permission) so that I can give this PR a good review :)

May I do that?

matrush:
Yeah, feel free to squash it. I'm okay with that :)

NSExceptional:
All done then :) Fixed some formatting and other stuff like that and threw in a fix for how numeric columns are sorted. Thanks for this!

PR #443, opened 2020-07-10
Add better debugging message for FLEXProperty and FLEXIvar

matrush:
Add better debugging message for `FLEXProperty` and `FLEXIvar`.

When the assertion failed, we want to know which class and which name are the problematic combinations.

### DIFF
--- a/Classes/Utility/Runtime/Objc/Reflection/FLEXIvar.m
+++ b/Classes/Utility/Runtime/Objc/Reflection/FLEXIvar.m
@@ -37,7 +37,8 @@
 }
 
 + (instancetype)named:(NSString *)name onClass:(Class)cls {
-    Ivar ivar = class_getInstanceVariable(cls, name.UTF8String);
+    Ivar _Nullable ivar = class_getInstanceVariable(cls, name.UTF8String);
+    NSAssert(ivar, @"Cannot find ivar with name %@ on class %@", name, cls);
     return [self ivar:ivar];
 }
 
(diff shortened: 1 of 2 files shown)


NSExceptional:
I intended those to be nullable. When are you running into a missing ivar?

matrush:
> I intended those to be nullable. When are you running into a missing ivar?

This is the iOS9 issue I’m trying to fix in the other PR. The ‘sharedApplication’ might not be found on UIApplication, thus I’m adding this debug message to help verify.

matrush:
Oh you are talking about the Ivar, I didn’t see the reasoning for that and didn’t running into issues after patch so I did it the same way as properties. If it’s intentional feel free to remove.

Also I think we should use _Nullable/_Nonnull specifies more broadly in the codebase in order to deal with some cases in a proper way, and to improve readability too :)

NSExceptional:
Agree! I have some free time right now, I'll try to sprinkle a bunch of nullability specifies around the runtime model classes with the fix for this

NSExceptional:
I actually don't have an iOS 9 device with me at the moment… can you debug it for me and see exactly where it's crashing?

If you enable the exception breakpoint you should be able to go up the stack trace and see the call with the metadata type and name. I bet it's the `sharedApplication` property but I just wanna be sure.

matrush:
> I actually don't have an iOS 9 device with me at the moment… can you debug it for me and see exactly where it's crashing?
> 
> If you enable the exception breakpoint you should be able to go up the stack trace and see the call with the metadata type and name. I bet it's the `sharedApplication` property but I just wanna be sure before I go fix it.

Haha, I don’t have one either, that’s why I’m adding the message and submitted that speculative fix... 😅

NSExceptional:
Gotcha, alright. I have a device at home as well but I'm not at my own apartment at the moment. I'll have that PR up in a bit for you to test

PR #444, opened 2020-07-11
Use UIApplication.class instead of UIApplication.flex_metaclass

matrush:
Use `UIApplication.class` instead of `UIApplication.flex_metaclass`

See https://github.com/Flipboard/FLEX/issues/424, this somehow has issue on iOS 9 somehow from my internal reports, so I think this can be a speculative fix.


NSExceptional:
This is intentional. Copying my reply from the other thread:

Adding shortcuts to the metaclass makes them appear when you explore the class itself. Adding them to the class makes them appear on instances. You can add class properties as shortcuts to either

If something is crashing iOS 9, we need to find out what or why

NSExceptional:
I bet the issue is that the property isn't actually a property under iOS 9, so we would need to add the property at runtime like we're doing with a bunch of UIView properties.

matrush:
Makes sense, can you give a pointer on where we are doing that for other properties?

NSExceptional:
Sure! Both haha. I'll mention you in the PR so you can see.

PR #445, opened 2020-07-11
Add sharedApplication property at runtime for iOS 9

NSExceptional:
cc @matrush


NSExceptional:
Did this fix it @matrush?

PR #449, opened 2020-07-15
Fix keyboard shortcut registration

maxolls:
### Issue
If the app adds custom keyboard shortcuts before FLEX sets default, the defaults will override the custom keys. For example, we use "t" to open our internal test settings (not FLEX), but that broke recently. Specifically, `[FLEXManager registerDefaultSimulatorShortcuts]` gets called after a `dispatch_async`, which gives the app enough time to register before.
### Fix
When setting the default FLEX keys, lets not allow overrides. This way, if the app sets custom keys at any time (event at `+load`), the app custom keys get the priority.


ryanolsonk:
Thanks @maxolls!

PR #454, opened 2020-07-20
Fix crash when sharing single file on iPad

TekYin:
Fixing the crash on sharing single file on iPad by setting the `shareSheet.popoverPresentationController.sourceView = sender`
error:
```swift
Terminating app due to uncaught exception 'NSGenericException', 
reason: 'UIPopoverPresentationController (<UIPopoverPresentationController: 0x7fc18ec54320>) should have a non-nil sourceView or barButtonItem set before the presentation occurs.'
```


NSExceptional:
Thanks!

badger200:
Wow! I have a number of show stopping crashes in major apps like Instagram due to this popoverpresentation bug. I'm extremely familiar with it only because I recognize seeing it so many times in my Cr4shed logs. It always causes Instagram to insta-crash. 😄

I can't believe it's that easy to fix, 1 line of code!?

(I use tweaks to force iPhone-only apps into full iPad mode. I used TonyKraft's PerfectFit tweak for years but recently switched to Spark's NoMoreSmallApps which works even better, and now AppPad which does the hat trick of enabling rotating, SlideOver, and SplitView for all apps!!!! You can even SlideOver iOS Settings.app!!)

PR #459, opened 2020-07-31
Resolve Xcode 12 linker error

cl7:
Address [Xcode 12 linker error](https://github.com/Flipboard/FLEX/issues/434), where `+load` method in two categories causing conflicts.

### DIFF
--- a/Classes/Manager/FLEXManager+Extensibility.h
+++ b/Classes/Manager/FLEXManager+Extensibility.h
@@ -59,6 +59,8 @@ NS_ASSUME_NONNULL_BEGIN
                                   action:(dispatch_block_t)action
                              description:(NSString *)description;
 
++ (void)loadExtensibility;
+
 @end
 
 NS_ASSUME_NONNULL_END
--- a/Classes/Manager/FLEXManager+Extensibility.m
+++ b/Classes/Manager/FLEXManager+Extensibility.m
@@ -170,7 +170,7 @@
     } description:@"Toggle file browser menu"];
 }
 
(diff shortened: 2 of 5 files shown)


NSExceptional:
Thanks! Looking at this now

NSExceptional:
This is strange; according to Mike Ash (who works on the Objc runtime), it is perfectly safe and defined behavior to implement `+load` in a category alongside the class itself, and in other categories:

> An interesting feature of +load is that it's special-cased by the runtime to be invoked in categories which implement it as well as the main class. This means that if you implement +load in a class and in a category on that class, both will be called. This probably goes against everything you know about how categories work, but that's because +load is not a normal method. This feature means that +load is an excellent place to do evil things like method swizzling.

From [his blog](https://www.mikeash.com/pyblog/friday-qa-2009-05-22-objective-c-class-loading-and-initialization.html)

I want to reach out to him about this before we merge this; there is still time for Xcode 12 to revert this behavior if it is indeed an accidental regression as I suspect it may be.

NSExceptional:
Mike [agrees](https://twitter.com/mikeash/status/1289280102442897408?s=21), could you file a Feedback and give him or me the number?

cl7:
@NSExceptional  Thanks! I have [filed this feedback 8228590](https://feedbackassistant.apple.com/feedback/8228590).

NSExceptional:
Thank you! Closing this for now, we can always reopen it if things don't change before the final release this fall. Don't delete your branch!

PR #461, opened 2020-08-19
Add foundation header to ActivityStreamSPI.

justinctlam:
The purpose of this change is to gate C++ code behind a cplusplus define so that FLEX can be made into a module for Swift code.  Swift modules do not allow for C++ code.  However this means if FLEX is used a Swift module, it will not have functionality of the excluded code. Ideally in the future, FLEX can be converted fully to support Swift without bridging headers.

### DIFF
--- a/Classes/GlobalStateExplorers/SystemLog/ActivityStreamAPI.h
+++ b/Classes/GlobalStateExplorers/SystemLog/ActivityStreamAPI.h
@@ -14,6 +14,8 @@
 #ifndef ActivityStreamSPI_h
 #define ActivityStreamSPI_h
 
+#include <Foundation/Foundation.h>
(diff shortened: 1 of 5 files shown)


justinctlam:
cc: @matrush

NSExceptional:
Bear with me, I don't know what this really fixes. Doesn't FLEX already work in Swift projects?

justinctlam:
Sorry about the original changes, I worked with a colleague to find a much better change.

justinctlam:
For some context, our company uses a different build setup internally, and we need all headers to be able to stand on their own. This file is using NS_ENUM, which is defined in Foundation, without importing it, so we're adding the import to resolve that.

PR #464, opened 2020-09-10
Swift Package Manager support

mstfy:
This pr adds spm support for FLEX.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ DerivedData
 /Example/Pods
 Podfile.lock
 IDEWorkspaceChecks.plist
+.swiftpm/
--- a/Classes/Core/Controllers/FLEXFilteringTableViewController.h
+++ b/Classes/Core/Controllers/FLEXFilteringTableViewController.h
@@ -25,7 +25,7 @@
 @property (nonatomic, copy) NSArray<FLEXTableViewSection *> *allSections;
 
 /// This computed property should filter \c allSections for assignment to \c sections
-@property (nonatomic, readonly) NSArray<FLEXTableViewSection *> *nonemptySections;
+@property (nonatomic, readonly, copy) NSArray<FLEXTableViewSection *> *nonemptySections;
 
 /// This should be able to re-initialize \c allSections
 - (NSArray<FLEXTableViewSection *> *)makeSections;
@@ -80,7 +80,7 @@
 /// if using \c self as the \c filterDelegate, as is the default.
 ///
 /// For example, the object explorer hides the description section when searching.
-@property (nonatomic, readonly) NSArray<FLEXTableViewSection *> *nonemptySections;
+@property (nonatomic, readonly, copy) NSArray<FLEXTableViewSection *> *nonemptySections;
 
 /// If using \c self as the \c filterDelegate, as is the default,
 /// subclasses should override to provide the sections for the table view.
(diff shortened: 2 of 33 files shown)


NSExceptional:
Hi,

I'm not familiar with how SPM works, but is there no better way to add header search paths than by manually specifying each directory in the project?

I'm not sure I want to merge this if that is truly what is required; its prone to breaking easily and makes refactoring difficult, and new additions tedious. And if this is not what is required, we need to do it the correct way

mstfy:
Unfortunately there is no way to add header search path recursively. Here is the [documentation](https://developer.apple.com/documentation/swift_packages/cxxsetting/3112674-headersearchpath). 

I checked some other objc projects that supports spm and found out they do similar thing. Either create symlink from all headers to `include` folder or specify all header search paths. Here are some example Package.swift files from other open source projects:

1 - [Realm](https://github.com/realm/realm-cocoa/blob/master/Package.swift)
2 - [TOCropViewController](https://github.com/TimOliver/TOCropViewController/blob/master/Package.swift)

NSExceptional:
According to NeoNacho on Twitter, you can use the `#file` variable to get the path of the manifest file, and from there you can recursively enumerate the directories of the project instead of hard-coding them. Can you try that?

I hope in the future they will just add recursive search paths

mstfy:
This works when you just build package. But when you add it as dependency to xcode project xcode copies `Package.swift` somewhere else and compiles it there. That makes impossible to reach source folder of package. Here is `Package.swift` 

```
import PackageDescription

let package = Package(
    name: "FLEX",
    products: [
        .library(name: "FLEX", targets: ["FLEX"])
    ],
    targets: [
        .target(
            name: "FLEX",
            path: "Classes",
            exclude: [
                "GlobalStateExplorers/SystemLog/LLVM_LICENSE.TXT",
                "Network/PonyDebugger/LICENSE",
                "Utility/APPLE_LICENSE",
                "GlobalStateExplorers/Keychain/SSKeychain_LICENSE",
                "Info.plist",
                "GlobalStateExplorers/DatabaseBrowser/LICENSE"
            ],
            cSettings: headerSearchPaths()
        )
    ]
)

import Foundation
func headerSearchPaths() -> [CSetting] {
    let root = #file.components(separatedBy: "/")
        .dropLast()
        .joined(separator: "/")
        .appending("/Classes")
    let rootUrl = URL(fileURLWithPath: root)

    let headerPaths = try! FileManager.default.subpathsOfDirectory(atPath: root)
        .compactMap { path -> CSetting? in
            let url = URL(fileURLWithPath: path, relativeTo: rootUrl)
            if (try? url.resourceValues(forKeys: [.isDirectoryKey]))?.isDirectory == true {
                return CSetting.headerSearchPath(path)
            } else {
                return nil
            }
        }

    return headerPaths
}
```

NSExceptional:
I see :/

I think I will take this issue to the forums, and if no one there has any suggestions in sight then we can work on merging this with the hard-coded paths.

canius:
Seams it will be missing extenstion funcs when import FLEX using spm. eg: FLEXManager.shared.registerGlobalEntry()

mdellagi77:
Hello there. 
Could you tell me if this is something coming in the short term? 
Our team would like to migrate to SPM in this is one of the last dependencies that are still not SPM compatible. 
Much obliged!

NSExceptional:
Closing this in favor of #482 which resolves an issue in this PR and has fewer merge conflicts to resolve. I apologize for taking so long to get to either of these PRs :(

Issue #470, opened 2020-10-05
Some way to disable loading of FLEX at runtime


lilyball:
I'd really like to be able to completely disable loading of FLEX at runtime, either with a launch argument (e.g. something that sets a volatile defaults key) or with an env variable.

The idea here is that I really don't need FLEX to be loading while running unit/UI tests in CI, especially given crashes like #469. It's meant to be an aid for exploring/debugging, not something that should be loaded when I'm running automated tests. But since building for testing is just doing a debug build, the test app includes FLEX.

In particular, what I'd like to do is disable fishhook, and anything else that runs at load time. The framework would still be linked into my app, so if I have code that programmatically shows it, that code will continue to link, but it's okay for FLEX to be nonfunctional if I actually try and show it with the runtime disabled (a quick hack might be for `showExplorer()` to simply do nothing and log an error if invoked with the runtime disabled, and of course the default simulator keyboard shortcuts shouldn't exist in this case though I assume disabling run-at-load behavior would do that already).

matrush:
I'm also looking at this, I think it can be achievable by introducing some C preprocessor so it can be opt-in or opt-out when building.

lilyball:
That would require me to pass a custom build setting flag to `xcodebuild` when building in CI, which is doable, but also prevents me from sharing any build artifacts with other builds (e.g. if I want to disable in unit testing but actually have it in UI testing for some reason, the two would have to independently rebuild the app).

I would much prefer a runtime check, so I can disable it without having to rebuild.

matrush:
I'm not entirely sure if that's achievable since the current loads are done in the `+ (void)load` class method. It is invoked whenever a class or category is added to the Objective-C runtime; It is unlikely you can do any call to other parts of `FLEX` to tweak that behavior before it loads. So I think the preprocessor way is probably the most obvious solution. Please correct me if I'm wrong.

lilyball:
I'm saying the `+load` methods should consult either an env var or `NSUserDefaults` to check for a predefined var/key whose presence signals "don't load FLEX". If this var/key is present, the `+load` method would then just exit without doing anything.

[...]

PR #471, opened 2020-10-05
Add macro to optionally disable FLEX loading at runtime

chrisczupak:
Adding the ability to disable FLEX from loading at runtime by adding the `DISABLE_FLEX_RUNTIME_LOAD` macro in project settings. 

This allows you to make runtime changes without conflicting with FLEX, and also aligns with a recent request that was made for this behavior: https://github.com/FLEXTool/FLEX/issues/470.


matrush:
@NSExceptional This looks to me, how do you think about this approach vs the one mentioned in #470 for using `NSUserDefaults`?

NSExceptional:
I'm sort of thinking we should accept both! Why not, right?

I'm going to take a deeper look at this this weekend. Not ready to merge it yet

NSExceptional:
I've reviewed the PR (and merged it with some changes I've made yesterday so that it is up to date with `master`) and this is actually not what we need in #470. I thought this was going to add a macro to check for a _runtime_ flag but all this does is determine at compile-time whether the constructors will run or not.

@chrisczupak, would you be okay with a dynamic solution like using an environment variable or an `NSUserDefaults` key? If so, I can close this and I will add that myself. If you need a compile-time macro though, let me know!

NSExceptional (commented):
I've reviewed the PR (and merged it with some changes I've made yesterday so that it is up to date with `master`) and this is actually not what we need in #470. I thought this was going to add a macro to check for a _runtime_ flag but all this does is determine at compile-time whether the constructors will run or not.

@chrisczupak, would you be okay with a dynamic solution like using an environment variable or an `NSUserDefaults` key? If so, I can close this and I will add that myself. If you need a compile-time macro though, let me know!

chrisczupak:
Hi, @NSExceptional! Sorry for the delay. I saw the change to [Add FLEX_EXIT_IF_NO_CTORS](Add FLEX_EXIT_IF_NO_CTORS) which looks good. However, I'd still like a compile time macro, so if we can move ahead with this PR as well that would be fantastic! Thank you!

NSExceptional:
Okay! The way I've set it up we won't be using this PR though, so no need to keep pushing commits :P

I did force push your branch to get it up to date with master btw

chrisczupak:
haha ok, thanks for the heads up!

NSExceptional:
Yeah 😆

NSExceptional:
Use `FLEX_DISABLE_CTORS` as of `4.2.0`

PR #473, opened 2020-10-06
Improve prettyJSONStringFromData

RaziPour1993:


NSExceptional:
Can you elaborate on what your changes do? Looks to me like it just makes this method return NULL if it can't prettify a JSON string, which is not what we want.

PR #476, opened 2020-10-23
Fix crash on `flex_withTintColor`

oleghnidets:
Hi,

There is a crash when you open Network History. Method unrecognized `flex_withTintColor`. I discovered that naming is incorrect in header and implementation (.m).

`Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '-[UIBarButtonItem flex_withTintColor:]: unrecognized selector sent to instance 0x7fafcfe6d1f0'`

Please, consider this MR. 

Thanks for your work!


NSExceptional:
Whoops, thank you

PR #482, opened 2020-11-09
Swift Package Manager support

RafaelFernandezAlv:
Hi, I add support for swift package manager for this awesome library.

I saw the PR #464 but this contains an error with the functions of extensions (you can't use it).

This PR solve this error and change the README.md for explain how to install with Swift Package Manager.

Bassically the support consist in add a new folder "Headers" with all ".h" headers of the library. This ".h" are a symlink to the reals in other paths and they are generated with the new script `create_headers.sh` in the root repo. The `Package.swift` use this Header folder as public headers of the library. Feel free to add some exceptions in the script if you want some of this ".h" as private headers.

Also, I added a sample target in the project with FLEX installed with Swift Package Manager

### DIFF
--- a/Classes/FLEX-Categories.h
+++ b/Classes/FLEX-Categories.h
@@ -6,17 +6,25 @@
 //  Copyright © 2020 FLEX Team. All rights reserved.
 //
 
+#import "UIBarButtonItem+FLEX.h"
+#import "CALayer+FLEX.h"
+#import "UIFont+FLEX.h"
+#import "UIGestureRecognizer+Blocks.h"
+#import "UIView+FLEX_Layout.h"
+#import "UIPasteboard+FLEX.h"
+#import "UIMenu+FLEX.h"
+#import "UITextField+Range.h"
 
 
-#import <FLEX/UIBarButtonItem+FLEX.h>
-#import <FLEX/CALayer+FLEX.h>
-#import <FLEX/UIFont+FLEX.h>
-#import <FLEX/UIGestureRecognizer+Blocks.h>
-#import <FLEX/UIPasteboard+FLEX.h>
-#import <FLEX/UIMenu+FLEX.h>
-#import <FLEX/UITextField+Range.h>
+#import "UIBarButtonItem+FLEX.h"
+#import "CALayer+FLEX.h"
+#import "UIFont+FLEX.h"
+#import "UIGestureRecognizer+Blocks.h"
+#import "UIPasteboard+FLEX.h"
+#import "UIMenu+FLEX.h"
+#import "UITextField+Range.h"
 
-#import <FLEX/NSObject+FLEX_Reflection.h>
-#import <FLEX/NSArray+FLEX.h>
-#import <FLEX/NSUserDefaults+FLEX.h>
-#import <FLEX/NSTimer+FLEX.h>
+#import "NSObject+FLEX_Reflection.h"
+#import "NSArray+FLEX.h"
+#import "NSUserDefaults+FLEX.h"
+#import "NSTimer+FLEX.h"
--- a/Classes/FLEX-Core.h
+++ b/Classes/FLEX-Core.h
@@ -6,18 +6,18 @@
 //  Copyright © 2020 FLEX Team. All rights reserved.
 //
 
-#import <FLEX/FLEXFilteringTableViewController.h>
-#import <FLEX/FLEXNavigationController.h>
-#import <FLEX/FLEXTableViewController.h>
-#import <FLEX/FLEXTableView.h>
+#import "FLEXFilteringTableViewController.h"
+#import "FLEXNavigationController.h"
+#import "FLEXTableViewController.h"
+#import "FLEXTableView.h"
 
-#import <FLEX/FLEXSingleRowSection.h>
-#import <FLEX/FLEXTableViewSection.h>
+#import "FLEXSingleRowSection.h"
+#import "FLEXTableViewSection.h"
 
-#import <FLEX/FLEXCodeFontCell.h>
-#import <FLEX/FLEXSubtitleTableViewCell.h>
-#import <FLEX/FLEXTableViewCell.h>
-#import <FLEX/FLEXMultilineTableViewCell.h>
-#import <FLEX/FLEXKeyValueTableViewCell.h>
+#import "FLEXCodeFontCell.h"
+#import "FLEXSubtitleTableViewCell.h"
+#import "FLEXTableViewCell.h"
+#import "FLEXMultilineTableViewCell.h"
+#import "FLEXKeyValueTableViewCell.h"
 
-#import <FLEX/FLEXScopeCarousel.h>
+#import "FLEXScopeCarousel.h"
--- a/Classes/FLEX-ObjectExploring.h
+++ b/Classes/FLEX-ObjectExploring.h
@@ -6,25 +6,25 @@
 //  Copyright © 2020 FLEX Team. All rights reserved.
 //
 
-#import <FLEX/FLEXObjectExplorerFactory.h>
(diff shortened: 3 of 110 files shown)


NSExceptional:
Do you mind if I edit this PR? I prefer merging one logical change at a time per PR, so I'd like to remove all the code changes that are not related to SPM support, even if it fixes warnings.

RafaelFernandezAlv:
Ok, no problem.

Thanks!

JackoPlane:
@NSExceptional Any update on supporting SPM?

NSExceptional:
@JackoPlane I actually forgot this PR existed; I will see to it ASAP as more and more people have been asking about it.

kravtsovguy:
@NSExceptional I ask you about it as you wished =)

Claes34:
Hey guys, any update on when this could be merged ? Thanks :D

NSExceptional:
Hey @RafaelFernandezAlv,

First of all, I'm sorry I took so long to review this!

I would like to make some small changes before I merge this; would you be okay with giving me write access temporarily?

Normally when you create a PR there is a checkbox to give maintainers write access (I believe just to the specific branch related to the PR, actually) but since the PR is already open, I think you just have to add me as a contributor to your fork

RafaelFernandezAlv:
Hi @NSExceptional. I give you write permissions to the fork. Tell me if you need anymore

NSExceptional:
@RafaelFernandezAlv I just noticed this, but why was it necessary to change the import syntax for SPM? We need to continue using `<FLEX/...>` syntax for everything else.

RafaelFernandezAlv:
I don't remember very well, but SPM keeps the folder hierarchy intact when resolving imports and modifying the imports was the only solution I could find.

NSExceptional:
@RafaelFernandezAlv Well, we can address that later I suppose as I've just run into bigger issues

The script you wrote links ALL headers in the `Headers` folder, which is NOT what we want. We only want to include the public headers. I have modified the script to do this. However, because of this:

> SPM keeps the folder hierarchy intact when resolving imports

FLEX itself will not build. This is really frustrating. This also goes back to my comments in the previous PR, where you must manually specify each folder in the header search paths, which is **HORRIBLE** 😫

I have updated the script to also output a list of `.headerSearchPath`s we can copy into Package.swift by hand, in the meantime.

Lastly, we need to use `.unsafeFlags` to silence warnings (if there is another way, I don't know what it is). This appears to make it impossible to use the package at all for some stupid reason. See: https://forums.swift.org/t/confused-by-unsafe-flags-being-disallowed-in-dependencies/27359/

Any ideas?

TheCoordinator:
Hey guys, does anybody have an update on this?

NSExceptional:
Not from me. Not sure where to go from here. Don't want to add _official_ SPM support without a way to silence warnings…

apps4everyone:
an easy way would be to support SPM support with binary target with pre compiled xcFramework.

TheCoordinator:
Agree with @apps4everyone this will be the best way for an application like this. I know Firebase is doing this now, would it be possible to follow what they have done in here too?

apps4everyone:
opened a PR: https://github.com/FLEXTool/FLEX/pull/550
Package.swift file:
https://github.com/FLEXTool/FLEX/pull/550/files#diff-f913940c58e8744a2af1c68b909bb6383e49007e6c5a12fb03104a9006ae677e

NSExceptional:
I definitely do not want to pre-compile FLEX for every release, I don't have time to do that. Would it be possible to include the Package.swift file in a sub folder so anyone can do that as needed?

alessaba:
Isn’t there free GitHub CI to auto compile?

TheCoordinator:
Yes! It can certainly be done simply on Github actions CI for free.

apps4everyone:
> I definitely do not want to pre-compile FLEX for every release, I don't have time to do that. Would it be possible to include the Package.swift file in a sub folder so anyone can do that as needed?

it is one command line with: `bundle exec fastlane exportXCFramework`
but I agree: this could be generated on github CI on release. But I am not an expert in this. Maybe this can work as an solution until someone else can help here?

glassomoss:
@apps4everyone I've tried this approach at our project, but unfortunately didn't get it to work on simulator. App just crashes when launched without Xcode. I've tried both adding it to another Swift Packet and directly to project. The error is 

`Dyld Error Message:
  Library not loaded: /Library/Frameworks/FLEX.framework/FLEX
`

I haven't tried it out on the device though. Xcode version is 13.1.

apps4everyone:
@glassomoss the problem is not the xcFramework it's the version 4.5.0, there was a fix adterwords but there was never a release @NSExceptional

apps4everyone:
https://github.com/FLEXTool/FLEX/pull/547

apps4everyone:
can you try master?

glassomoss:
@NSExceptional oh I see, thank you. I guess I could try right now, but I haven't really built any xcframework before. Could you give me a clue where to start?)

apps4everyone:
@glassomoss hopefully this will help:
https://github.com/FLEXTool/FLEX/pull/551#issuecomment-903293295

glassomoss:
@apps4everyone Thanks! 

Long story short: I failed at literally every step. I wasn't able to install bundle, it just kept failing with random errors, so I ended up installing fastlane with brew. Then I ran into https://github.com/bielikb/xcframeworks/issues/10 with fastlane and had to do build xcframework manually https://www.appcoda.com/xcframework/. 

As I described my initial problem was app just crashing when launched without Xcode. I created an all new project and FLEX appeared to be working just fine. So yes, this solution works. Great job!

My conclusion is that the crucial difference between my project and the all-new one is having multiple (static?) packages vs having only FLEX. 

As for my first project, I've been struggling for hours and I ended up connecting FLEX to another Swift Package rather than directly to the whole project. So that might be the solution for these problems which potential users should be aware of. 

However, I'm not sure what exactly was the problem in connecting FLEX directly. I guess there might be better solution.

apps4everyone:
@glassomoss thx for the report, sorry for the bundle problem. Yes the real problem is the MISSING release after the fix https://github.com/FLEXTool/FLEX/pull/547 @NSExceptional

glassomoss:
None of my issues were related to that problem

glassomoss:
For those who wondering, I was able to exclude FLEX from the project by changing `EXCLUDED_SOURCE_FILE_NAMES` to `Flex*` in Release

NSExceptional:
@RafaelFernandezAlv I ended up writing a bunch of scripts to make it easier to maintain the package manifest, and I also went with a pretty custom manifest compared to yours, so it was just less work to push my own changes instead of merging this PR. You are still the author on some of the commits though I think. Thank you for your help everyone.

Debrief: SPM doesn't allow use of unsafe flags in Xcode projects, so I just removed them. FLEX will just have to generate warnings in SPM. The "public headers" situation is still shitty, but this will do for now.

NSExceptional:
Oh, also @apps4everyone @glassomoss I released `4.6.1`

apps4everyone:
@NSExceptional there is no 4.6.1 release visible only a tag but this tag branch has no Package.swift

NSExceptional:
All you need is a tag for a Cocoapods release. As for SPM, just use `master` for now.

apps4everyone:
thx for the information, master is NO valid option for us because if something like https://github.com/FLEXTool/FLEX/pull/547 is going to master the app will crash, we have to pin a specific variosion.

NSExceptional:
Then use a specific commit hash

glassomoss:
> thx for the information, master is NO valid option for us because if something like #547 is going to master the app will crash, we have to pin a specific version.

@apps4everyone I've managed to fix this issue, as I mentioned in one of my previous comments

PR #493, opened 2020-12-13
Fix view controller preview image issue

matrush:
This issue was introduced in https://github.com/FLEXTool/FLEX/commit/547bfbaec0c0910ee4c135722ff02c27d93826b6, there is a `!` missed in the boolean check, so fixing it here.

I also consolidated some logic around this, also apply the same thing to the `CALayer` as well.

### DIFF
--- a/Classes/ObjectExplorers/Sections/Shortcuts/FLEXLayerShortcuts.m
+++ b/Classes/ObjectExplorers/Sections/Shortcuts/FLEXLayerShortcuts.m
@@ -15,11 +15,11 @@
 + (instancetype)forObject:(CALayer *)layer {
     return [self forObject:layer additionalRows:@[
         [FLEXActionShortcut title:@"Preview Image" subtitle:nil
-            viewer:^UIViewController *(id layer) {
+            viewer:^UIViewController *(CALayer *layer) {
                 return [FLEXImagePreviewViewController previewForLayer:layer];
             }
-            accessoryType:^UITableViewCellAccessoryType(id layer) {
-                return UITableViewCellAccessoryDisclosureIndicator;
+            accessoryType:^UITableViewCellAccessoryType(CALayer *layer) {
+                return CGRectIsEmpty(layer.bounds) ? 0 : UITableViewCellAccessoryDisclosureIndicator;
             }
         ]
     ]];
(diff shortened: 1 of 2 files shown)


Classes/ObjectExplorers/Sections/Shortcuts/FLEXLayerShortcuts.m
```diff
@@ -15,11 +15,11 @@ @implementation FLEXLayerShortcuts
 + (instancetype)forObject:(CALayer *)layer {
     return [self forObject:layer additionalRows:@[
         [FLEXActionShortcut title:@"Preview Image" subtitle:nil
-            viewer:^UIViewController *(id layer) {
+            viewer:^UIViewController *(CALayer *layer) {
```
NSExceptional, line 18: Does the compiler not warn or complain about this? If not, nice!

Classes/ObjectExplorers/Sections/Shortcuts/FLEXViewShortcuts.m
```diff
@@ -73,13 +73,15 @@ + (instancetype)forObject:(UIView *)view {
                 return controller ? UITableViewCellAccessoryDisclosureIndicator : 0;
             }
         ],
-        [FLEXActionShortcut title:@"Preview Image" subtitle:nil
-            viewer:^UIViewController *(id view) {
+        [FLEXActionShortcut title:@"Preview Image" subtitle:^NSString *(UIView *view) {
+                return !CGRectIsEmpty(view.bounds) ? @"" : @"Unavailable because view.bounds is CGRectZero";
```
NSExceptional, line 77: Thanks for this detail! Can we re-word it though? How about this?
```suggestion
                return !CGRectIsEmpty(view.bounds) ? @"" : @"Unavailable with empty bounds";
```

NSExceptional (changes requested):

NSExceptional:
Whoops, good catch!

Classes/ObjectExplorers/Sections/Shortcuts/FLEXLayerShortcuts.m
matrush: It doesn't warn on compiling :)

PR #497, opened 2020-12-29
Fix navbar buttons tap handling

magauran:
Navigation bar buttons didn't work on iOS 11-12. The bug was reproduced only on a real device.

Tested on:
- iPhone 6 Plus, iOS 11.0.3
- iPhone 11 Pro, iOS 14.2

### DIFF
--- a/Classes/Core/Controllers/FLEXNavigationController.m
+++ b/Classes/Core/Controllers/FLEXNavigationController.m
@@ -36,10 +36,13 @@
     self.waitingToAddTab = YES;
     
     // Add gesture to reveal toolbar if hidden
-    self.navigationBar.userInteractionEnabled = YES;
-    [self.navigationBar addGestureRecognizer:[[UITapGestureRecognizer alloc]
+    UITapGestureRecognizer *navbarTapGesture = [[UITapGestureRecognizer alloc]
         initWithTarget:self action:@selector(handleNavigationBarTap:)
-    ]];
+    ];
+    
+    // Don't cancel touches to work around bug on versions of iOS prior to 13
+    navbarTapGesture.cancelsTouchesInView = NO;
+    [self.navigationBar addGestureRecognizer:navbarTapGesture];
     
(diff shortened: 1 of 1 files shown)


NSExceptional:
You say you tested on an 11 Pro on iOS 14, are you saying you tested there and it did work on iOS 14? Or that it does work on iOS 14?

magauran:
It did work on iOS 14. I tested just in case to make sure it still works and I didn't break anything.
I've only encountered this bug on iOS 11 and iOS 12.

Classes/Core/Controllers/FLEXNavigationController.m
```diff
@@ -36,10 +36,10 @@ - (void)viewDidLoad {
...
     
     // Add gesture to reveal toolbar if hidden
-    self.navigationBar.userInteractionEnabled = YES;
-    [self.navigationBar addGestureRecognizer:[[UITapGestureRecognizer alloc]
-        initWithTarget:self action:@selector(handleNavigationBarTap:)
-    ]];
+    __auto_type navigationBarTapGestureRecognizer = [[UITapGestureRecognizer alloc] initWithTarget:self
+                                                                                            action:@selector(handleNavigationBarTap:)];
+    navigationBarTapGestureRecognizer.cancelsTouchesInView = NO;
+    [self.navigationBar addGestureRecognizer:navigationBarTapGestureRecognizer];
```
NSExceptional, line 42:
```suggestion
    UITapGestureRecognizer *navbarTapGesture = [[UITapGestureRecognizer alloc]
        initWithTarget:self action:@selector(handleNavigationBarTap:)
    ];
    
    // Don't cancel touches to work around bug on versions of iOS prior to 13
    navigationBarTapGestureRecognizer.cancelsTouchesInView = NO;
    [self.navigationBar addGestureRecognizer:navbarTapGesture];
```

NSExceptional (changes requested):

PR #501, opened 2021-01-20
Fix a build issue caused by changes in FLEXObjectListViewController

matrush:
Fix build issue caused by changes in `FLEXObjectListViewController`.

Some changes were added in https://github.com/FLEXTool/FLEX/commit/e63ea4bbfff4200e4d371b1fd86448975162898d, it works fine by default, but if some advanced compile flags were enabled, it will start failing.

So here we added the missing `static` descriptor and removed an unused function.

### DIFF
--- a/Classes/GlobalStateExplorers/FLEXObjectListViewController.m
+++ b/Classes/GlobalStateExplorers/FLEXObjectListViewController.m
@@ -29,20 +29,12 @@ typedef NS_ENUM(NSUInteger, FLEXObjectReferenceSection) {
     FLEXObjectReferenceSectionCount
 };
 
-NSArray<NSString *> * FLEXObjectReferenceSectionTitles() {
-    return @[
-        @"", @"AutoLayout", @"Key-Value Observing", @"FLEX"
-    ];
-}
-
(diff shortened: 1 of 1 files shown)


NSExceptional:
Oh crap, it's unused because I meant to call it in `defaultSectionTitles` inside that loop! Can you replace the call to the default predicate method with a call to the function you removed? (And add it back of course; probably better to reset back to `master` first and force push your PR)

matrush:
@NSExceptional Haha sure, I didn't look at the logic closely. I will update the changes now.

matrush:
After taking another look, the `FLEXTitleForObjectReferenceSection` function is unused, but the `FLEXObjectReferenceSectionTitles` is used. It seems the current logic is still correct, maybe you can take another look at it?

NSExceptional:
I will get back to this in the morning 😅

matrush:
I think this is probably what you mean :) I updated the change. Feel free to take a look tomorrow, it's not urgent.

NSExceptional:
That looks good! But now that you mention it, I don't know why I added those things as functions at all in the first place… your way looks cleaner, with the `defaultSectionTitles` method.

I got rid of the functions entirely, and will merge it in!

NSExceptional (approved):

PR #521, opened 2021-03-22
Fix viewing keychain crash.

betteray:
Fix crashing when viewing keychain that may contains NSData.


NSExceptional:
Looks like you're fixing two different things here. Can we break the second change into a separate PR?

betteray:
This PR only fix the viewing keychain crash now.

PR #522, opened 2021-03-23
Fix viewing keychain item crash.

betteray:


NSExceptional:
Thank you for helping me identify the crash! Unfortunately this is not the best way to fix it, so I have replaced your fix with the actual fix so that you will still be credited as a contributor.

PR #529, opened 2021-04-05
Fix: Correctly quote table names

### LINKED ISSUE #524: Unquoted table names causes issue when table has reserved sql keyword
### Environment
- Platform+version: **iOS 14.x**
- FLEX version: **4.4.1** 

### Bug Report

https://github.com/FLEXTool/FLEX/blob/258ec8697bf9c703390b0994899d94e289406c63/Classes/GlobalStateExplorers/DatabaseBrowser/FLEXSQLiteDatabaseManager.m#L117-L119

The name of my table is `order`, causing the following statement from my sql logging:
`near "order": syntax error in "SELECT * FROM order"`

The correct statement should read `SELECT * FROM "order"`

mgray88:
Sqlite is often smart enough to not need quotes for identifiers, unless you want to use a keyword as a name. Then you need to quote it

https://www.sqlite.org/lang_keywords.html

Fixes #524

### DIFF
--- a/Classes/GlobalStateExplorers/DatabaseBrowser/FLEXSQLiteDatabaseManager.m
+++ b/Classes/GlobalStateExplorers/DatabaseBrowser/FLEXSQLiteDatabaseManager.m
@@ -114,9 +114,7 @@ static NSString * const QUERY_TABLENAMES = @"SELECT name FROM sqlite_master WHER
 }
 
 - (NSArray<NSArray *> *)queryAllDataInTable:(NSString *)tableName {
-    return [self executeStatement:[@"SELECT * FROM "
-        stringByAppendingString:tableName
-    ]].rows ?: @[];
+    return [self executeStatement:[NSString stringWithFormat:@"SELECT * FROM \"%@\"", tableName]].rows ?: @[];
 }
 
 - (FLEXSQLResult *)executeStatement:(NSString *)sql {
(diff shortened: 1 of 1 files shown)


NSExceptional:
Thanks! Could you do me a favor and break the command string out into it's own variable on a new line?

mgray88:
@NSExceptional Like just `tableName` similar to how it was before?

```
return [self executeStatement:[NSString stringWithFormat:@"SELECT * FROM \"%@\"",
    tableName
]].rows ?: @[];
```

NSExceptional:
Nah, hang on, let me do a formal code suggestion

Classes/GlobalStateExplorers/DatabaseBrowser/FLEXSQLiteDatabaseManager.m
```diff
@@ -114,9 +114,7 @@ - (NSInteger)lastRowID {
 }
 
 - (NSArray<NSArray *> *)queryAllDataInTable:(NSString *)tableName {
-    return [self executeStatement:[@"SELECT * FROM "
-        stringByAppendingString:tableName
-    ]].rows ?: @[];
+    return [self executeStatement:[NSString stringWithFormat:@"SELECT * FROM \"%@\"", tableName]].rows ?: @[];
```
NSExceptional, line 117:
```suggestion
    NSString *command = [NSString stringWithFormat:@"SELECT * FROM \"%@\"", tableName];
    return [self executeStatement:command].rows ?: @[];
```

mgray88: Gotcha. That was my other interpretation :)

PR #530, opened 2021-04-08
Fix build: The project file is missing some files to import.

BB9z:
These files are added in https://github.com/FLEXTool/FLEX/commit/c37270e6acd5c13f7170a671d62faab7230fa857, but not included in the project file.

I found this problem when installing with Carthage. I suggest that a new revised version should be released.

Temporary workaround is to specify an older version:

```
github "FLEXTool/FLEX" "4.4.0"
```


NSExceptional:
Heh, whoops. Thanks for bringing this to my attention.

Unfortunately I think I'll just have to close this and fix it with my own commit, because upon further inspection, there was another pair of files missing as well, + Xcode decided to do an upgrade check and asked me to change some build settings. All that combined with the fact that Xcode decided to shuffle the files around in your project file for no apparent reason means merging them would be a huge headache.

I did credit you in the commit message though :) Thanks again

PR #531, opened 2021-04-10
Make it possible to access swift ivars

LNSSPsd:


NSExceptional:
I am currently working on a supplemental library called Reflex that will add full support for swift reflection to FLEX :D

Does this code actually work though? I was under the impression swift ivars returned an offset of 0

LNSSPsd:
I've tested it on Music,Calculator and an app downloaded from App Store,and it seems that this function works correctly for most swift ivars and I haven't found one that returns 0 as its offset.

NSExceptional:
Ah, it returns 0 as the size, that's what I was thinking of.

Question: do you find this of any real use for anything but accessing objet pointers? It gives everything back to us as NSValue which isn't very useful for most things, except integers or floats I guess. Would you be okay if I merged this with just support for looking at objects instead of any arbitrary ivar? What are your thoughts here?

LNSSPsd:
Maybe we can add an option that disabled by default to force show the descriptions of those ivars, and also a context menu to force unwrap them. btw maybe it's also okay that just replacing the descriptions of those object-pointer-like ivars to a tip like "`Click to force unwrap`" since they have no default action?

NSExceptional:
Yeah I like the last idea

LNSSPsd:
Tried to implement it on the last commit.

NSExceptional:
Closing this PR as I have begun official work on fleshing this sort of functionality out with [FLEXTool/Reflex](https://github.com/FLEXTool/Reflex)

PR #534, opened 2021-04-18
Fixed the frame error when showing FLEX for the first time

Saafo:
When showing FLEX for the first time when the app is in split view (not in full screen) on iPad, FLEX will layout itself as the width of the whole screen:

(you may need to disabled the auto-opening FLEX code in `SceneDelegate.swift` in FLEXample project to reproduce this case)

Simulator Screen Shot - iPad Pro (11-inch) (2nd generation) - 2021-04-18 at 20 32 56

This pr fixed this bug.

BTW, I'm not familiar with objc, so feel free to edit this pr if needed~

### DIFF
--- a/Classes/Manager/FLEXManager.m
+++ b/Classes/Manager/FLEXManager.m
@@ -73,7 +73,7 @@
     if (@available(iOS 13.0, *)) {
         // Only look for a new scene if we don't have one
         if (!flex.windowScene) {
-            flex.windowScene = FLEXUtility.activeScene;
+            flex.windowScene = FLEXUtility.appKeyWindow.windowScene;
         }
     }
 #endif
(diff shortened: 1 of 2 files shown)


Saafo:
Another issue confused me is that, in the case of opening two scenes of my app in the foreground, both of the two scenes are marked as `active window` by system. However, when I opened the flex by the button in the second window in foreground, it will appear in the first foreground active window because you use

```objc
flex.windowScene = FLEXUtility.activeScene;
```

in `FLEXManager.m`. And the activeScene is just to get the first active scene by using for-loop.

In this case, I prefer using `FLEXUtility.appKeyWindow.windowScene` rather than `FLEXUtility.activeScene`.

I can open another pr for this issue if needed.

Saafo:
Fixed the second issue in the last commit above

NSExceptional:
Thanks so much for this! Looks good to me!

PR #536, opened 2021-04-23
Change pushed time of tree vc to avoid empty page.

suixudongi8:
It pushes nil tree before initialization sometimes.
Delay the push time to avoid this.

### DIFF
--- a/Classes/ViewHierarchy/FLEXHierarchyViewController.m
+++ b/Classes/ViewHierarchy/FLEXHierarchyViewController.m
@@ -83,6 +83,9 @@ typedef NS_ENUM(NSUInteger, FLEXHierarchyViewMode) {
 
     // Start of in tree view
     _mode = FLEXHierarchyViewModeTree;
+}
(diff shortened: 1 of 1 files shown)


suixudongi8:
Please have a review @bdotdub @drodriguez @robinsonrc @dlo

NSExceptional:
So it crashes sometimes is what you're saying? I need to review my code but I wouldn't  expect an entire view controller to be nil like that, so that sounds like a bug that needs fixing instead

suixudongi8:
viewDidLoad will be hooked by other app module to do some custom config. It will be triggered wrongly before treeVC is inited.
So it's more safe to show treeVC in viewWillAppear stage.

NSExceptional:
`treeViewController` is *always* initialized ***before*** `viewDidLoad`, so something else is definitely going on. I do not want to merge this to cover up some other bug your code. I'm glad to help you figure out what that may be, if you want

PR #540, opened 2021-06-23
Add logic to flush WAL data to sql file after file opening

LesykMelnychuk:
Logic to make data in sql file actual, moving it from cache.

### DIFF
--- a/Classes/GlobalStateExplorers/DatabaseBrowser/FLEXSQLiteDatabaseManager.m
+++ b/Classes/GlobalStateExplorers/DatabaseBrowser/FLEXSQLiteDatabaseManager.m
@@ -57,6 +57,9 @@ static NSString * const QUERY_TABLENAMES = @"SELECT name FROM sqlite_master WHER
 
     if (err != SQLITE_OK) {
         return [self storeErrorForLastTask:@"Open"];
+    } else {
+        NSString *sql = @"VACUUM;"; // To flesh WAL file into sqlite
+        [self executeStatement:sql];
     }
     
     return YES;
--- a/Classes/GlobalStateExplorers/DatabaseBrowser/FLEXSQLiteDatabaseManager.m
+++ b/Classes/GlobalStateExplorers/DatabaseBrowser/FLEXSQLiteDatabaseManager.m
@@ -57,10 +57,10 @@ static NSString * const QUERY_TABLENAMES = @"SELECT name FROM sqlite_master WHER
 
     if (err != SQLITE_OK) {
         return [self storeErrorForLastTask:@"Open"];
-    } else {
-        NSString *sql = @"VACUUM;"; // To flesh WAL file into sqlite
-        [self executeStatement:sql];
     }
+    
+    NSString *sql = @"VACUUM;"; // To flesh WAL file into sqlite
+    [self executeStatement:sql];
     
     return YES;
 }
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,0 +1,3 @@
+# These are supported funding model platforms
+
+github: [NSExceptional]


NSExceptional:
Can I ask what this does? I don't know much about SQLite. I couldn't understand your description, I'm sorry

LesykMelnychuk:
> Can I ask what this does? I don't know much about SQLite. I couldn't understand your description, I'm sorry

It flashed WAL file data into sqlite file, its done for multithreading access. If not do this - data  made by app, can be missed in sqlite file

Classes/GlobalStateExplorers/DatabaseBrowser/FLEXSQLiteDatabaseManager.m
```diff
@@ -59,6 +59,9 @@ - (BOOL)open {
         return [self storeErrorForLastTask:@"Open"];
     }
     
+    NSString *sql = @"VACUUM;"; // To flesh WAL file into sqlite
+    [self executeStatement:sql];
```
matrush, line 63: Can you check the `PRAGMA schema.journal_mode` first and only do this when we are in the WAL mode?

matrush (changes requested):

Classes/GlobalStateExplorers/DatabaseBrowser/FLEXSQLiteDatabaseManager.m
LesykMelnychuk: Hello `matrush`, how can I do this in Objective C ?

matrush: It is basically the same as what you are doing already, just use `[self executeStatement:sql]` to execute the `PRAGMA schema.journal_mode`, and then check the result.

Checkout [this](https://github.com/FLEXTool/FLEX/blob/master/Classes/GlobalStateExplorers/DatabaseBrowser/FLEXSQLiteDatabaseManager.m#L108) as an example.

NSExceptional: @matrush I can add this, what do I use in place of `schema`?

NSExceptional:
@LesykMelnychuk Is this something that needs to be done after every mutating query when the database is in WAL mode? Or just once when the database is opened?

LesykMelnychuk:
> @LesykMelnychuk Is this something that needs to be done after every mutating query when the database is in WAL mode? Or just once when the database is opened?

Each time

Classes/GlobalStateExplorers/DatabaseBrowser/FLEXSQLiteDatabaseManager.m
matrush: Yeah I think that might just work, maybe try and see what it returns?

badger200:
@NSExceptional I would be cautious about implementing this request. Www.sqlite.org has a very good FAQ page on WAL (Write-Ahead Log) explaining all misconceptions and potential pitfalls. 

Most importantly: the WAL **_IS_** part of the database. It is NOT a mere “log file” in any traditional sense. SQLite tells you straight up: deleting the WAL will destroy recent data created that hasn’t yet been committed. A LOT of intelligent engineering has gone into the WAL system for maximum disk performance. There is no legit circumstance where “failing” to VACUUM would result in recent data not being displayed. Many iOS system apps/services literally *never* commit their WAL; if you’re jailbroken or on Mac OS X just do a ```
stat whatever.sqlite*``` from terminal and you’ll see when the WAL was created. Often it is years ago when you first bought your device: proving the WAL has never been committed (because VACUUM removes the WAL file upon completion.)

My highest concern is large databases such as every device’s /User/Media/PhotoData/photos.sqlite (or CPL/cloudphotos.db) which on my iPad with 150,000 photos is 5.5GB, can have a massive WAL exceeding 1GB which takes a long time and a ton of disk thrashing to commit; a needless performance penalty to suffer arbitrarily. iOS schedules system daemons like assetsd (the Photos.app/Camera Roll/DB primary daemon) to perform these types of IO and CPU intensive tasks during overnight charging when the device has been inactive for >2h etc. 

You can view these rules by running ```launchctl print system/com.apple.assetsd``` or use com.apple.photoanalysisd for the daemon that examines photos with AI for categorizing contents like “Dog, Summer, September 2013, Beach, Jane Doe, John Doe, Miami, Florida” etc. Another good one is com.apple.cache_delete_app_container_caches. (Apple spells this one inconsistently across the system, might need to double check as I know the service name does _not_ match the daemon executable name.)

You can issue ```launchctl list system``` to get a list of all services to query. Run ```launchctl list help``` if I got that wrong. But my first example (assetsd) is definitely correct as I’ve done it so many times I know it by heart. 

Conclusion: SQLite developers would say it’s officially inappropriate to arbitrarily vacuum every time you access a database, especially due to the huge performance task you may inadvertently commence. When my iPad Pro would begin one of those 1GB photos.sqlite-wal commits, my device would lag to the point of almost unusable for several minutes. 

It would be fine to have an optional checkbox for VACUUM or a setting “Always VACUUM SQLite DBs upon close?” but that too seems flawed; one could say there should be additional options like ```
PRAGMA integrity_check;``` etc. 

What Filza does is to have a SQLite command textbox atop the DB display; you can type VACUUM; yourself if you want. But coincidentally I’m pretty sure Filza actually does indeed automatically VACUUM every DB you open whether you changed it or not; this has been frustrating for me because sometimes the .sqlite file still has the original time stamp of when it was created, but the -wal is always dated current day (last activity) - but upon VACUUM, the merge occurs and the resulting sole .sqlite files dated present time, losing/destroying the original created date forever. 

Usually not a big deal but I’ve used such dates in many instances:
- forensic analysis
- troubleshooting looking for clues for when a long-standing problem first began occurring 
- when I open a sqlite and then want to know if it’s identical to a copy I made recently

NSExceptional:
@badger200 I agree, we shouldn't add any code that "touches" the file or modifies it in any way without explicit user consent. 

> There is no legit circumstance where “failing” to VACUUM would result in recent data not being displayed.

Does this mean that opening the sqlite file and running queries on it will pull data from the adjacent WAL file automatically? If that's the case, then it would seem you _don't_ ever need to `VACUUM` it, indeed, and this PR should just be closed. Can you confirm this for me?

NSExceptional:
Closing this as I have reason to believe flushing is entirely unnecessary. I can reopen this if someone replies with additional information

PR #542, opened 2021-07-14
Add a toolbar button to delete all records in a table

hossamghareeb:
**Problem:**

At some cases, we need to delete all records in a table to replicate a specific scenario. The only way to do this right now is to run SQL query from tables list. 

**This PR:**

Adds a toolbar button when you open a table to delete all records. Now it's easier than writing a SQL query.

WIP: Add an option to delete a specific row.

https://user-images.githubusercontent.com/1477248/125656863-b51e1af1-7461-4265-bbd5-d42792535572.mp4

### DIFF
--- a/Classes/GlobalStateExplorers/DatabaseBrowser/FLEXTableContentViewController.h
+++ b/Classes/GlobalStateExplorers/DatabaseBrowser/FLEXTableContentViewController.h
@@ -7,10 +7,19 @@
 //
 
 #import <UIKit/UIKit.h>
+#import "FLEXDatabaseManager.h"
+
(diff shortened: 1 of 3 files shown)


NSExceptional:
I like this, do you mind if I make a few minor edits to the copy in the alert?

hossamghareeb:
> I like this, do you mind if I make a few minor edits to the copy in the alert?

@NSExceptional not at all. Go ahead

hossamghareeb:
@NSExceptional any updates on this? I would like to land these changes soon. Thanks

NSExceptional:
Sorry for the delay, been busy

PR #543, opened 2021-07-14
Option to delete a row

hossamghareeb:
Adding the option to delete a specific row. Due to the absence of knowledge of primary keys, I decided to use rowid in SQL to delete rows. Rowids will be injected along with rows and columns. 

This PR depends on #542 

https://user-images.githubusercontent.com/1477248/125669245-7f480303-7a94-4776-8bd7-317cc7035bf5.mp4


NSExceptional:
Thanks!

PR #548, opened 2021-08-05
Add availability check for iOS 11+ SFSafariViewController properties

matrush:
Add availability check for iOS 11+ SFSafariViewController properties.

The [`configuration`](https://developer.apple.com/documentation/safariservices/sfsafariviewcontroller/configuration) and [`dismissButtonStyle`](https://developer.apple.com/documentation/safariservices/sfsafariviewcontroller/dismissButtonStyle) properties are both iOS 11+ only. They were added in https://github.com/FLEXTool/FLEX/commit/bbaa85bdbffe2b42cf06cbda43f5ec6393ce99a9. Let's add availability check to make sure there is no runtime assertion on certain OS versions.

The error I got was:
```
NSInternalInconsistencyException at +[FLEXProperty named:onClass:] in FLEXProperty.m:49: Cannot find property with name configuration on class SFSafariViewController
```

### DIFF
--- a/Classes/ObjectExplorers/Sections/Shortcuts/FLEXShortcutsFactory+Defaults.m
+++ b/Classes/ObjectExplorers/Sections/Shortcuts/FLEXShortcutsFactory+Defaults.m
@@ -377,11 +377,16 @@
     
     if (SafariVC) {
         self.append.properties(@[
-            @"delegate", @"configuration", @"dismissButtonStyle"
+            @"delegate"
         ]).forClass(SafariVC);
         if (@available(iOS 10.0, *)) {
             self.append.properties(@[
                 @"preferredBarTintColor", @"preferredControlTintColor"
+            ]).forClass(SafariVC);
+        }
+        if (@available(iOS 11.0, *)) {
+            self.append.properties(@[
+                @"configuration", @"dismissButtonStyle"
             ]).forClass(SafariVC);
         }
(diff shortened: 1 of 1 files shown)


Classes/ObjectExplorers/Sections/Shortcuts/FLEXShortcutsFactory+Defaults.m
```diff
@@ -25,7 +25,7 @@ + (void)load { FLEX_EXIT_IF_NO_CTORS()
     FLEXRuntimeUtilityTryAddObjectProperty(
         2, sharedApplication, UIApplication.flex_metaclass, UIApplication, PropertyKey(ReadOnly)
     );
-
```
matrush, line 28: Well, my editor auto removes all these spaces, feel free to revert these changes if you prefer to do so.

NSExceptional: Yeah, I'll do this in the morning 😅

For what it's worth, if you use Git Tower, it's really easy to just commit the functional changes and then discard all the leftover whitespace changes! I haven't used other apps so idk how well they handle line-by-line changes, and I know it's a pain if you're just using the command line

NSExceptional:
Whoops, sorry for the delay

Issue #554, opened 2021-09-29
sqlite database viewer is broken


AprilC:
### Environment
- Platform+version: **iOS 14 + SQLite 3.28.0**
- FLEX version: **4.5.0**

### Bug Report

sqlite database viewer is broken on iOS due to `PRAGMA table_info('%@')` not working on certain sqlite versions such as 3.28.0. All table in database viewer are empty. The issue is this statement for column names isn't returning any rows.
See more discussion in #555

PR #555, opened 2021-09-29
Fix SQLite database viewer

AprilC:
Fix bug  #554: sqlite database viewer is broken

SQLite database viewer is broken on iOS. All table in database viewer are empty. The issue is this statement for column names isn't returning any rows. We don't know the reason yet. But even if we can get the column names from the current api, we will run the query for 3 times. With this change, we added a new API: queryTable. This API only needs to query the table once and get everything.

### DIFF
--- a/Classes/GlobalStateExplorers/DatabaseBrowser/FLEXSQLiteDatabaseManager.m
+++ b/Classes/GlobalStateExplorers/DatabaseBrowser/FLEXSQLiteDatabaseManager.m
@@ -32,7 +32,7 @@ static NSString * const QUERY_TABLENAMES = @"SELECT name FROM sqlite_master WHER
     if (self) {
         self.path = path;
     }
-    
+
     return self;
 }
 
@@ -44,7 +44,7 @@ static NSString * const QUERY_TABLENAMES = @"SELECT name FROM sqlite_master WHER
     if (self.db) {
         return YES;
     }
-    
+
     int err = sqlite3_open(self.path.UTF8String, &_db);
 
 #if SQLITE_HAS_CODEC
@@ -58,18 +58,18 @@ static NSString * const QUERY_TABLENAMES = @"SELECT name FROM sqlite_master WHER
     if (err != SQLITE_OK) {
         return [self storeErrorForLastTask:@"Open"];
     }
-    
+
     return YES;
 }
-    
+
 - (BOOL)close {
     if (!self.db) {
         return YES;
     }
-    
+
     int  rc;
     BOOL retry, triedFinalizingOpenStatements = NO;
-    
+
     do {
         retry = NO;
         rc    = sqlite3_close(_db);
@@ -89,7 +89,7 @@ static NSString * const QUERY_TABLENAMES = @"SELECT name FROM sqlite_master WHER
             return NO;
         }
     } while (retry);
-    
+
     self.db = nil;
     return YES;
 }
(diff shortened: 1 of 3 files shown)


Classes/GlobalStateExplorers/DatabaseBrowser/FLEXSQLiteDatabaseManager.m
```diff
@@ -267,34 +272,34 @@ - (FLEXSQLResult *)errorResult:(NSString *)description {
...
 }
 
 - (id)objectForColumnIndex:(int)columnIdx stmt:(sqlite3_stmt*)stmt {
     int columnType = sqlite3_column_type(stmt, columnIdx);
-    
+
     switch (columnType) {
         case SQLITE_INTEGER:
-            return @(sqlite3_column_int64(stmt, columnIdx)).stringValue;
+            return [NSString stringWithFormat:@"%lld", sqlite3_column_int64(stmt, columnIdx)];
```
matrush, line 284: Is this change necessary?

Classes/GlobalStateExplorers/DatabaseBrowser/FLEXSQLiteDatabaseManager.m
```diff
@@ -267,34 +272,34 @@ - (FLEXSQLResult *)errorResult:(NSString *)description {
...
         case SQLITE_FLOAT:
-            return  @(sqlite3_column_double(stmt, columnIdx)).stringValue;
+            return [NSString stringWithFormat:@"%f", sqlite3_column_double(stmt, columnIdx)];
```
matrush, line 286: Same for this one, is it necessary?

Classes/GlobalStateExplorers/DatabaseBrowser/FLEXSQLiteDatabaseManager.m
AprilC: in some cases, i got an exception for this. it's safe to change to stringWithFormat

Classes/GlobalStateExplorers/DatabaseBrowser/FLEXSQLiteDatabaseManager.m
AprilC: see above.

matrush:
I took another look at this.
It seems this is an issue for certain sqlite versions (at least on 3.28 that I've tested).
See here for more reports:
https://sqlite.org/forum/info/2ca63507adffb34e
Here is another one (the last reply said it's patched in the latest version)
https://sqlite.org/forum/info/0e55af1e0b1823f7

So I agree given the `PRAGMA table_info(...)` has issues on certain versions, we should try to avoid using it if possible.

Back to the current commit, it won't work as the SQLite specific logic is brought to `FLEXTableListViewController` which is suppose to be generic as it supports other DBs as well. So I think we have to do some extra work here. The performance turning is good but it might not be straightforward to support that in `FLEXRealmDatabaseManager`.

Classes/GlobalStateExplorers/DatabaseBrowser/FLEXTableListViewController.m
```diff
@@ -67,64 +67,71 @@ - (void)viewDidLoad {
...
+
     self.tables.selectionHandler = ^(FLEXTableListViewController *host, NSString *tableName) {
-        NSArray *rows = [host.dbm queryAllDataInTable:tableName];
-        NSArray *columns = [host.dbm queryAllColumnsOfTable:tableName];
-        NSArray *rowIDs = [host.dbm queryRowIDsInTable:tableName];
-        UIViewController *resultsScreen = [FLEXTableContentViewController
-            columns:columns rows:rows rowIDs:rowIDs tableName:tableName database:host.dbm
-        ];
-        [host.navigationController pushViewController:resultsScreen animated:YES];
+        FLEXSQLResult *result = [host.dbm queryTable:tableName];
```
matrush, line 72: `FLEXTableListViewController` supports more than one db. You cannot have `FLEXSQLResult` dependencies here.

matrush (changes requested):
See my above comment for more details but it cannot be merged as is.

NSExceptional:
I'll see if I can find time to look at this this week 🤔 thanks @matrush snd OP

Classes/GlobalStateExplorers/DatabaseBrowser/FLEXSQLiteDatabaseManager.m
NSExceptional: Do you have a crash log for that? I'm highly skeptical that could ever result in an exception/crash; `@()` should accept any return value and properly wrap it in `NSNumber`

NSExceptional:
Closing in favor of the other PR

PR #556, opened 2021-09-30
Fix realm database viewer crash

### LINKED ISSUE #533: How to add for bazel swift project
I'm trying to add flex to a project but when I drag the folder classes or the project to the folder as explained in the manual section it doesn't seem to work when I press "f". 

Is there some other way to add the project to a bazel built project or some step with manually adding it that I'm not doing correctly?

skytoup:
Fix #533


matrush (approved):
Since `queryRowIDsInTable` is optional method, this makes sense to me.

PR #566, opened 2021-11-05
Fix CLANG_WARN_STRICT_PROTOTYPES/-Wstrict-prototypes issues

mattrobmattrob:
This allows `FLEX` to be linked statically to a binary that has `-Wstrict-prototypes` enabled (for example in CocoaPods without `use_frameworks!` since `spec.compiler_flags` aren't propagated).

Changes:
- Apply `void` to all the empty function/block declarations that don't take arguments.
- Apply `SEL`/`UIImage *` to the others that actually take arguments.
- Remove `CLANG_WARN_STRICT_PROTOTYPES = NO` since the default is enabled.


matrush:
This looks good to me, thanks for resolving the warning :)

PR #567, opened 2021-11-05
Fix FLEXMacros header visibility in xcodeproj

mattrobmattrob:
Make the visibility of `FLEXMacros.h` (in the repo's Xcode project) match the visibility declared in `FLEX.podspec`.

### DIFF
--- a/Classes/GlobalStateExplorers/SystemLog/FLEXOSLogController.m
+++ b/Classes/GlobalStateExplorers/SystemLog/FLEXOSLogController.m
@@ -170,7 +170,7 @@ static uint8_t (*OSLogGetType)(void *);
             // Get log message text
             const char *messageText = OSLogCopyFormattedMessage(log_message);
             // https://github.com/limneos/oslog/issues/1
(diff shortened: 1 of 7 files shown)


NSExceptional:
Are there any other headers with Private visibility? Was this the only one without correct visibility?

mattrobmattrob:
> Are there any other headers with Private visibility? Was this the only one without correct visibility?

I'm not sure, @NSExceptional. This was just fixing the following warning when building the project in this repo:

>.../FLEX/Classes/FLEX.h:25:1: Umbrella header for module 'FLEX' does not include header 'FLEXMacros.h'

Which is hard to compare against the CocoaPods inclusion because of how CocoaPods maps the headers:

Screen Shot 2021-11-07 at 3 54 23 PM

NSExceptional:
I think I fixed this. Could you re-clone the repo and confirm for me?

Closing for now, can reopen if the issue persists. Fixed it myself because there were _many_ files with this issue, not just this one.

PR #573, opened 2021-11-18
Add fallback column query support when PRAGMA table_info is not working

### LINKED ISSUE #554: sqlite database viewer is broken

matrush:
Add fallback column query support when `PRAGMA table_info` is not working.

This is my attempt to fix #554, which is a less intrusive way comparing to #555. More detail about the bug itself can be found in #555.

The change I did here are two parts:
1. Fallback to use a slightly hackier solution when the `PRAGMA table_info` is not working.
2. Do not consider 0 row result as an `UPDATE/INSERT/DELETE`, it's common that our result has 0 rows.

### DIFF
--- a/Classes/GlobalStateExplorers/DatabaseBrowser/FLEXSQLResult.m
+++ b/Classes/GlobalStateExplorers/DatabaseBrowser/FLEXSQLResult.m
@@ -13,7 +13,7 @@
 @synthesize keyedRows = _keyedRows;
 
 + (instancetype)message:(NSString *)message {
-    return [[self alloc] initWithmessage:message columns:nil rows:nil];
+    return [[self alloc] initWithMessage:message columns:nil rows:nil];
 }
 
 + (instancetype)error:(NSString *)message {
@@ -23,12 +23,12 @@
 }
 
 + (instancetype)columns:(NSArray<NSString *> *)columnNames rows:(NSArray<NSArray<NSString *> *> *)rowData {
-    return [[self alloc] initWithmessage:nil columns:columnNames rows:rowData];
+    return [[self alloc] initWithMessage:nil columns:columnNames rows:rowData];
 }
 
-- (id)initWithmessage:(NSString *)message columns:(NSArray *)columns rows:(NSArray<NSArray *> *)rows {
+- (instancetype)initWithMessage:(NSString *)message columns:(NSArray<NSString *> *)columns rows:(NSArray<NSArray<NSString *> *> *)rows {
     NSParameterAssert(message || (columns && rows));
-    NSParameterAssert(columns.count == rows.firstObject.count);
+    NSParameterAssert(rows.count == 0 || columns.count == rows.firstObject.count);
     
     self = [super init];
     if (self) {
--- a/Classes/GlobalStateExplorers/DatabaseBrowser/FLEXSQLiteDatabaseManager.m
+++ b/Classes/GlobalStateExplorers/DatabaseBrowser/FLEXSQLiteDatabaseManager.m
@@ -158,7 +158,7 @@ static NSString * const QUERY_TABLENAMES = @"SELECT name FROM sqlite_master WHER
             return self.lastResult;
         }
         
-        // Grab columns
+        // Grab columns (columnCount will be 0 for insert/update/delete) 
         int columnCount = sqlite3_column_count(pstmt);
         NSArray<NSString *> *columns = [NSArray flex_forEachUpTo:columnCount map:^id(NSUInteger i) {
             return @(sqlite3_column_name(pstmt, (int)i));
@@ -176,8 +176,9 @@ static NSString * const QUERY_TABLENAMES = @"SELECT name FROM sqlite_master WHER
         }
         
         if (status == SQLITE_DONE) {
+            // columnCount will be 0 for insert/update/delete
             if (rows.count || columnCount > 0) {
-                // We selected some rows
+                // We executed a SELECT query
                 result = _lastResult = [FLEXSQLResult columns:columns rows:rows];
             } else {
                 // We executed a query like INSERT, UDPATE, or DELETE
--- a/Classes/GlobalStateExplorers/DatabaseBrowser/FLEXSQLiteDatabaseManager.m
+++ b/Classes/GlobalStateExplorers/DatabaseBrowser/FLEXSQLiteDatabaseManager.m
@@ -107,10 +107,15 @@ static NSString * const QUERY_TABLENAMES = @"SELECT name FROM sqlite_master WHER
 - (NSArray<NSString *> *)queryAllColumnsOfTable:(NSString *)tableName {
     NSString *sql = [NSString stringWithFormat:@"PRAGMA table_info('%@')",tableName];
     FLEXSQLResult *results = [self executeStatement:sql];
-
-    return [results.keyedRows flex_mapped:^id(NSDictionary *column, NSUInteger idx) {
-        return column[@"name"];
-    }] ?: @[];
+    // https://github.com/FLEXTool/FLEX/issues/554
+    if (results.keyedRows.count == 0) {
+        NSString *command = [NSString stringWithFormat:@"SELECT * FROM \"%@\" where 0=1", tableName];
+        return [self executeStatement:command].columns ?: @[];
+    } else {
+        return [results.keyedRows flex_mapped:^id(NSDictionary *column, NSUInteger idx) {
+            return column[@"name"];
+        }] ?: @[];
+    }
 }
 
 - (NSArray<NSArray *> *)queryAllDataInTable:(NSString *)tableName {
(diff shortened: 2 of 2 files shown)


NSExceptional:
Ping me when you're done adding commits and I'll take a look at it! 😄

matrush:
@NSExceptional I think this would be it for now. Feel free to take a look :)

Classes/GlobalStateExplorers/DatabaseBrowser/FLEXSQLResult.m
```diff
@@ -23,12 +23,12 @@ + (instancetype)error:(NSString *)message {
 }
 
 + (instancetype)columns:(NSArray<NSString *> *)columnNames rows:(NSArray<NSArray<NSString *> *> *)rowData {
-    return [[self alloc] initWithmessage:nil columns:columnNames rows:rowData];
+    return [[self alloc] initWithMessage:nil columns:columnNames rows:rowData];
 }
 
-- (id)initWithmessage:(NSString *)message columns:(NSArray *)columns rows:(NSArray<NSArray *> *)rows {
+- (instancetype)initWithMessage:(NSString *)message columns:(NSArray<NSString *> *)columns rows:(NSArray<NSArray<NSString *> *> *)rows {
```
NSExceptional, line 29: Splendid!

Classes/GlobalStateExplorers/DatabaseBrowser/FLEXSQLResult.m
```diff
@@ -13,7 +13,7 @@ @implementation FLEXSQLResult
 @synthesize keyedRows = _keyedRows;
 
 + (instancetype)message:(NSString *)message {
-    return [[self alloc] initWithmessage:message columns:nil rows:nil];
+    return [[self alloc] initWithMessage:message columns:nil rows:nil];
```
NSExceptional, line 16: Good catch with these

Classes/GlobalStateExplorers/DatabaseBrowser/FLEXSQLiteDatabaseManager.m
```diff
@@ -107,10 +107,15 @@ - (NSInteger)lastRowID {
 - (NSArray<NSString *> *)queryAllColumnsOfTable:(NSString *)tableName {
     NSString *sql = [NSString stringWithFormat:@"PRAGMA table_info('%@')",tableName];
     FLEXSQLResult *results = [self executeStatement:sql];
-
-    return [results.keyedRows flex_mapped:^id(NSDictionary *column, NSUInteger idx) {
-        return column[@"name"];
-    }] ?: @[];
+    // https://github.com/FLEXTool/FLEX/issues/554
+    if (results.keyedRows.count == 0) {
+        NSString *command = [NSString stringWithFormat:@"SELECT * FROM \"%@\" where 0=1", tableName];
```
NSExceptional, line 112: Can I ask why you went with this instead of the approach documented in the links you posted?

```sql
SELECT * FROM pragma_table_info('sometable')
```
https://sqlite.org/forum/info/0e55af1e0b1823f7

Classes/GlobalStateExplorers/DatabaseBrowser/FLEXSQLiteDatabaseManager.m
```diff
@@ -164,7 +169,7 @@ - (FLEXSQLResult *)executeStatement:(NSString *)sql arguments:(NSDictionary *)ar
         }
         
         if (status == SQLITE_DONE) {
-            if (rows.count) {
+            if (rows.count || columnCount > 0) {
```
NSExceptional, line 172: `columnCount` is usually going to be non-zero here, this is going to cause the logic in the `else` to not execute
```suggestion
            if (rows.count) {
```

NSExceptional (commented):
Awaiting conversations to be resolved

Classes/GlobalStateExplorers/DatabaseBrowser/FLEXSQLiteDatabaseManager.m
matrush: For unknown reason, `SELECT * FROM pragma_table_info('sometable')` doesn't work per my testing. Since it's an ugly fallback, whatever works is the best :)

Classes/GlobalStateExplorers/DatabaseBrowser/FLEXSQLiteDatabaseManager.m
matrush: Is that true? My understanding was for `SELECT` statements it will have a positive count, for update/delete/insert it returns 0?
If that's not true then I think you are right.

Classes/GlobalStateExplorers/DatabaseBrowser/FLEXSQLiteDatabaseManager.m
matrush: Per my testing for some reason the `SELECT * FROM pragma_table_info('sometable')` query doesn't work. So I switched to use the other working method.

NSExceptional: Which iOS/Xcode version is this present in and can it be reproduced in the simulator? I want to toy around with it but I can't reproduce it

Classes/GlobalStateExplorers/DatabaseBrowser/FLEXSQLiteDatabaseManager.m
NSExceptional: Whoops, you are correct! I'm gonna add that as a comment for my future self

NSExceptional:
Force pushed because I rebased your fork on top of `FLEXTool/master`.

I notice you always "merge" your local copy; you should consider using rebase instead when you're updating your local repo with the remote branches, since I have "merge" disabled for this repo and never want to create "merge" commits 😅

Anyway, I'm ready to merge this except that I want to know how to reproduce this so I can test your fix to see it in action.

matrush:
Can you try testing against this combination: Platform+version: iOS 14 + SQLite 3.28.0
I think this is where I found the query is not working.
You can find 3.28 here: https://www.sqlite.org/changes.html

NSExceptional:
iOS comes with sqlite; are you bundling a different version of it with your application?

matrush:
> iOS comes with sqlite; are you bundling a different version of it with your application?

Yes, we have selectively using a custom version of SQLite to be consistent on all iOS versions. Also maybe certain version of iOS bundled SQLite also hit the same issue too.

NSExceptional:
Oof. Is SQLite available as a Cocoapod? How can I try this myself

matrush:
> 

Maybe try this:
https://github.com/clemensg/sqlite3pod

NSExceptional:
Is there anything special I need to do to make this work with FLEX since FLEX links the system one?

matrush:
> Is there anything special I need to do to make this work with FLEX since FLEX links the system one?

Not super familiar but I think maybe remove [this](https://github.com/FLEXTool/FLEX/blob/master/FLEX.podspec#L37) temporarily and use a custom linked sqlite library would work?

NSExceptional:
@matrush I was about to try out `sqlite3pod` but I just noticed they only use up to 3.27, 🆘

NSExceptional:
@matrush I would like to use the official query and fallback to your workaround instead of using your workaround as the default. I have this code on my machine locally but I would like to verify it works before I push it to your branch and merge this PR. Any help in getting me a local app I can test this in is appreciated!

matrush:
@NSExceptional 
I'm okay with the current logic you put out there:

`PRAGMA table_info('%@')` is the default
then it's
`SELECT * FROM pragma_table_info('%@')`
then it's
`SELECT * FROM \"%@\" where 0=1`.

https://github.com/clemensg/sqlite3pod/blob/master/sqlite3.podspec
This has been updated to 3.37.2 so I think you can test it now?

matrush:
Since I'm updating my internal version. I had the chance to test this again.

This is the snippet I'm using:
```
- (NSArray<NSString *> *)queryAllColumnsOfTable:(NSString *)tableName {
    NSString *sql = [NSString stringWithFormat:@"PRAGMA table_info('%@')",tableName];
    FLEXSQLResult *results = [self executeStatement:sql];
    NSLog(@"SQLite version: %s", SQLITE_VERSION);
    NSLog(@"Result 1 row: %@, column: %@", results.keyedRows, results.columns);
    // https://github.com/FLEXTool/FLEX/issues/554
    if (!results.keyedRows.count) {
        sql = [NSString stringWithFormat:@"SELECT * FROM pragma_table_info('%@')", tableName];
        results = [self executeStatement:sql];
        NSLog(@"Result 2 row: %@, column: %@", results.keyedRows, results.columns);

        // Fallback to empty query
        if (!results.keyedRows.count) {
            sql = [NSString stringWithFormat:@"SELECT * FROM \"%@\" where 0=1", tableName];
            results = [self executeStatement:sql];
            NSLog(@"Result 3 row: %@, column: %@", results.keyedRows, results.columns);
            return results.columns ?: @[];
        }
    }

    return [results.keyedRows flex_mapped:^id(NSDictionary *column, NSUInteger idx) {
        return column[@"name"];
    }] ?: @[];
}
```

The result:
```
2022-03-02 20:30:03.989249-0800 Messenger[85453:294580] SQLite version: 3.36.0
2022-03-02 20:30:03.989390-0800 Messenger[85453:294580] Result 1 row: (null), column: (null)
2022-03-02 20:30:05.828039-0800 Messenger[85453:294580] Result 2 row: (null), column: (null)
2022-03-02 20:30:05.828368-0800 Messenger[85453:294580] Result 3 row: (
), column: (
    "sort_order",
    id,
    "name",
)
```

You can see only the result 3 has valid column outputs.

matrush:
@NSExceptional Let's merge this? I have the proof attached in the above that the changes are working as intended.

NSExceptional:
Wow, your last reply was a month ago? I'm so sorry, I must have missed it 😭

Yes, good to merge! Thanks for this!

PR #575, opened 2021-11-23
Adding an option to add a row in DB table

hossamghareeb:
Added a new toolbar button to insert a new row. Tapping on the button will show an alert to write the values as a comma separated string which will be added in an INSERT statement to be executed.

https://user-images.githubusercontent.com/1477248/143076971-3c2f2b93-85e2-4e4f-ab22-741ed05dc078.mp4

### DIFF
--- a/Classes/GlobalStateExplorers/DatabaseBrowser/FLEXTableContentViewController.m
+++ b/Classes/GlobalStateExplorers/DatabaseBrowser/FLEXTableContentViewController.m
@@ -282,11 +282,15 @@
 
 - (void)addPressed {
     [FLEXAlert makeAlert:^(FLEXAlert *make) {
-        make.title(@"Insert a new row");
-        make.message(@"Type comma separated values to execute the INSERT statement.");
-        make.textField(@"Example: 5, 'John Smith', 14,...");
-        make.button(@"Run").handler(^(NSArray<NSString *> *strings) {
-            NSString *statement = [NSString stringWithFormat:@"INSERT INTO %@ VALUES (%@)", self.tableName, strings[0]];
+        make.title(@"Add a New Row");
+        make.message(@"Comma separate values to use in an INSERT statement.\n\n");
+        make.message(@"INSERT INTO [table] VALUES (your_input)");
+        make.textField(@"5, 'John Smith', 14,...");
+        make.button(@"Insert").handler(^(NSArray<NSString *> *strings) {
+            NSString *statement = [NSString stringWithFormat:
+                @"INSERT INTO %@ VALUES (%@)", self.tableName, strings[0]
+            ];
+
             [self executeStatementAndShowResult:statement completion:^(BOOL success) {
                 if (success) {
                     [self reloadTableDataFromDB];
--- a/Classes/GlobalStateExplorers/DatabaseBrowser/FLEXTableContentViewController.m
+++ b/Classes/GlobalStateExplorers/DatabaseBrowser/FLEXTableContentViewController.m
@@ -165,7 +165,7 @@
         
         // Option to delete row
         BOOL hasRowID = self.rows.count && row < self.rows.count;
-        if (hasRowID && self.databaseManager) {
+        if (hasRowID && self.canRefresh) {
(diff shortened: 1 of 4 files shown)


Classes/GlobalStateExplorers/DatabaseBrowser/FLEXTableContentViewController.m
```diff
@@ -260,5 +279,18 @@ - (void)executeStatementAndShowResult:(NSString *)statement completion:(void (^_
     } showFrom:self];
 }
 
+- (void)reloadTableDataFromDB {
+    if (self.databaseManager == nil) {
+        return;
+    }
+    NSArray *rows = [self.databaseManager queryAllDataInTable:self.tableName];
+    NSArray *rowIDs = nil;
```
matrush, line 287: `NSArray<NSString *> *`

matrush, line 287: Please add type generics to the NSArray while applicable
e.g. `NSArray<NSString *> *`

Classes/GlobalStateExplorers/DatabaseBrowser/FLEXTableContentViewController.m
```diff
@@ -132,7 +132,7 @@ - (void)multiColumnTableView:(FLEXMultiColumnTableView *)tableView didSelectRow:
         
         // Option to delete row
         BOOL hasRowID = self.rows.count && row < self.rows.count;
-        if (hasRowID) {
+        if (hasRowID && self.databaseManager) {
```
matrush, line 135: why do we need this change?

hossamghareeb: To make the option to delete a row only available if databaseManager is not null. It's null if you open the content screen after executing custom query from table screen.

matrush: It feels that we should figure out why it is `nil` and ensure it's always nonnull? Do you know why it will be `nil` in the scenario you mentioned?

Classes/GlobalStateExplorers/DatabaseBrowser/FLEXTableContentViewController.m
```diff
@@ -260,5 +279,18 @@ - (void)executeStatementAndShowResult:(NSString *)statement completion:(void (^_
     } showFrom:self];
 }
 
+- (void)reloadTableDataFromDB {
+    if (self.databaseManager == nil) {
```
matrush, line 283: same, was wondering if we can remove this after we ensure it's non-null

Classes/GlobalStateExplorers/DatabaseBrowser/FLEXTableContentViewController.m
NSExceptional: It is `nil` when the screen is the result of a custom query instead of tapping on a table.

I think I did this on purpose with the idea in mind that these screens should be immutable? There is a similar check already in `setupToolbarItems` in this file.

But that's silly now that I think about it. Better to just make it `nonnull` like @matrush said

NSExceptional: That said, `tableName` is `nonnull` and it _should_ be `nullable`… I think I need to do some refactoring here.

Classes/GlobalStateExplorers/DatabaseBrowser/FLEXTableContentViewController.m
NSExceptional: Without breaking this out into two separate classes (one for immutable tables and one for mutable tables) the best I can do is to wrap it in a property `self.canRefresh` which checks if we have `self.databaseManager` and `self.tableName`

NSExceptional:
Any thoughts on my changes before I merge this?

hossamghareeb:
@NSExceptional LGTM

PR #577, opened 2021-12-02
Adding an option to pin/unpin tables

hossamghareeb:
**Problem**:
Dealing with a huge list of tables is a headache, every time I have to scroll or search for the table I want to open. 

**Solution:**
Introduce a way to pin/unpin tables by long pressing on a table and divide the tables list to two sections pinned and unpinned

**How it works:**

https://user-images.githubusercontent.com/1477248/144338393-2c375d7f-97b1-499f-934c-d2c3e90a02bf.mov


NSExceptional:
I'm gonna be honest, this seems like a pretty niche feature 😅

Same with the other PR for adding new rows, but to a lesser extent.

If you wish, you're welcome to use some of the code from this repo and make an extension for FLEX that either adds a new section to the globals list or uses the objc runtime to modify and add behavior to this view controller.

hossamghareeb:
@NSExceptional I understand your opinion because you didn't experience the need of this feature. In apps that are heavily data driven and depend fully on DB to mimic the app state, they will have around for sure hundreds of tables where we depend fully on DB to debug as it is our source of truth. It's another superpower to add to this awesome debugging tool.

Thanks

NSExceptional:
What if instead we added a "recent tables" feature? It could keep the tables you've recently visited at the top, and it would not persist this data between launches

hossamghareeb:
@NSExceptional I really liked the recent tables features and it will make it easier I believe. I'm still in favour of persisting the data between launches. Usually while working with large number of tables, each team cares about a specific set of tables. So finding this set always at top in testing and debugging would make their life easier. Let me know what do you think?

hossamghareeb:
@NSExceptional Hey Tanner, could you have a look at my recent comment? wdyt?

PR #578, opened 2021-12-05
Remove extra whitespaces for whitespaces only lines

matrush:
Remove extra whitespaces for whitespaces only lines.

This is super annoying as the Xcode and VSCode both trims the spaces while editing.
I think doing a batch update to remove all the unneeded spaces is more future-proof. So here it is.

This is the command I have used:
```
find . -name "*" -exec sed -i '' 's/^[[:blank:]]*$//' {} \;
```

In the future we could consider setting up clang-format which should be able to do this automatically.


NSExceptional:
Thank you! I'm gonna leave this here for bit because I'm working on a feature that's going to have a lot of merge conflicts if I try to merge it after this 😅

matrush:
> Thank you! I'm gonna leave this here for bit because I'm working on a feature that's going to have a lot of merge conflicts if I try to merge it after this 😅

NP. I can redo it anytime. It's not in a hurry. Just let me know when you are ready :)

matrush:
@NSExceptional I reran it on the latest master. Let's try to get this in this time?

NSExceptional:
I've given this some additional though, and while I thought I was on-board with merging it some day, I think I may have changed my mind… here's why:

- I have Xcode set to _not_ trim whitespace-only lines—I'm guessing this isn't the default, but since I haven't touched this setting in years and it's off…
- it is apparently the setting I use in all of my projects, and
- I am reluctant to change that setting, because I like it personally, and
- this PR makes it so that I will have to choose between [changing my setting and reformatting all of my other repos, lest I introduce inconsistent whitespace formatting as I write new code,] or [making sure I trim the whitespace lines manually in FLEX going forward<sup>1</sup>]
- Merging this will result in so much headache when I finally go to merge in some of my old branches, probably fair to say more headache than you get<sup>2</sup> by discarding whitespace-only changes in your PRs ;P (well, if you're not using a git client that makes it easy to discard individual line changes, this may actually be pretty tedious—but Tower makes this easy)
- It will also cause merge conflicts on all the other PRs that are still up
- It will _wreck_ the git-blame history

<sup>1</sup> We should probably just set up a linter to format the code consistently instead of doing something like this by hand and leaving open the possibility of inconsistent formatting being committed later on
<sup>2</sup> While I have no idea what the default setting is, I noticed you're the only regular contributor who regularly gets caught up with lots of whitespace-only changes that I ask you to discard… O_o

For these reasons, I'm going to say that we probably shouldn't merge this at all :/

_I don't know if this helps since you probably edit this project mostly in Xcode, but in VS Code you can change settings on a per-folder basis and make it _not_ trim whitespace-only lines in this repo_

matrush:
Screen Shot 2022-03-03 at 2 29 31 AM
I think my Xcode setting is the same with you. It's probably the default setting of VSCode. I can opt-out for this folder. It's just typically it's uncommon to see people have whitespaces only lines in open source libraries. I'm don't feel strongly though, since at the end of the day the linter will probably just take care of this.

PR #582, opened 2022-03-03
Check rowIDs instead of rowIDs.count when validating input

matrush:
The row array could always be empty `@[]` for an empty db instead of `nil`. We shouldn't check `.count` which will cause crashes.


NSExceptional:
It's safe to send messages to `nil` and it's guaranteed to return 0, what crash is this avoiding?

Anyway, I think this was intentional to avoid displaying empty result sets. Do you want to do that? I don't feel strongly either way, so this is good to merge if that's what you want it to do now 🥸👍🏻

matrush:
`NSParameterAssert(all || none);` This is failing when you have an empty dataset. It's legit to have a table without any data in it, avoid displaying should be causing crashes in the first place :)

PR #583, opened 2022-03-03
Fix unused-property-ivar warning for bytesReceived

matrush:
`bytesReceived` is defined as a property
```
@property (nonatomic) NSInteger bytesReceived;
```
If we turn on `-Wunused-property-ivar`, this warning will fire because the `_bytesReceived` is not used at all.
```
/FLEX/src/Classes/Network/FLEXMITMDataSource.m:35:1: error: ivar '_bytesReceived' which backs the property is not referenced in this property's accessor [-Werror,-Wunused-property-ivar]
- (NSInteger)bytesReceived {
^
/FLEX/src/Classes/Network/FLEXMITMDataSource.h:25:33: note: property declared here
@property (nonatomic) NSInteger bytesReceived;
                                ^
1 error generated.
```

I understand that `_filteredBytesReceived` is equivalent to `_bytesReceived` if not filtered, but it's not super intuitive. So let's just do an explicit check and make sure to use the ivar to shut up the warning.

### DIFF
--- a/Classes/Network/FLEXMITMDataSource.m
+++ b/Classes/Network/FLEXMITMDataSource.m
@@ -29,11 +29,11 @@
 }
 
 - (NSArray *)transactions {
-    return _filteredTransactions;
+    return self.isFiltered ? _filteredTransactions : _allTransactions;
 }
 
 - (NSInteger)bytesReceived {
-    return _filteredBytesReceived;
+    return self.isFiltered ? _filteredBytesReceived : _bytesReceived;
 }
 
 - (void)reloadByteCounts {


NSExceptional:
Oof, I don't see that I'm even using `totalBytesReceived` _anywhere_… I think I wrote that class in a rush and forgot to tidy up some loose ends.

I think the `filtered___` properties are supposed to be equal to `bytesReceived` and `transactions`, so I should probably get rid of one of them. For example, I don't think we need this if statement at all, just the bottom half, because `bytesReceived` is already the same as `filteredBytesReceived`:

https://github.com/FLEXTool/FLEX/blob/2983649cdb00b418ba054f6bd41615813e03c0e9/Classes/Network/FLEXNetworkMITMViewController.m#L296-L305

I'm not exactly sure which one to get rid of, but I'll get on this. Thanks for bringing it to my attention ;P

matrush:
@NSExceptional Yeah, I don't know much about the actual logic here. Just trying to fix the build error :) Maybe you can follow up to clean it up further.

NSExceptional:
Gotcha, will do. Closing this for now, feel free to open an issue instead if you want to be aware of when I fix it!

matrush:
@NSExceptional Reopening this since this is blocking me updating FLEX with `-Wunused-property-ivar` turned on. This can be easily reproduced and it's easy to understand. I feel we can land my changes for now and you can follow up to clean it up further.

NSExceptional:
I'll see if I can get to this by the end of the day

NSExceptional:
20 days later, I finally got around to this… sorry 😅 Fixed

PR #592, opened 2022-03-31
Fix nil FLEXHTTPTransaction in recordRequestWillBeSentWithRequestID

### LINKED ISSUE #574: Crash in `FLEXNetworkRecorder`
### Environment
- Platform+version: **iOS 13** 
- FLEX version: **4.6.1** 

### Bug Report

0 | libobjc.A.dylib | _objc_retain
-- | -- | --
1 | CoreFoundation | _cow_copy_storage
2 | CoreFoundation | _cow_mutate_slow
3 | CoreFoundation | -[__NSArrayM insertObject:atIndex:]
4 | FLEX | __85-[FLEXNetworkRecorder recordRequestWillBeSentWithRequestID:request:redirectResponse:]_block_invoke
5 | libdispatch.dylib | _dispatch_call_block_and_release
6 | libdispatch.dylib | _dispatch_client_callout
7 | libdispatch.dylib | _dispatch_lane_serial_drain
8 | libdispatch.dylib | _dispatch_lane_invoke
9 | libdispatch.dylib | _dispatch_workloop_worker_thread
10 | libsystem_pthread.dylib | _pthread_wqthread
11 | libsystem_pthread.dylib | _start_wqthread

</div>

image

matrush:
The transaction could be `nil`, we have checks in other callbacks like `recordResponseReceivedWithRequestID` and `recordDataReceivedWithRequestID`. So we should have it here as well to avoid crash.

This should fix #574 as the stack trace is the same.

### DIFF
--- a/Classes/Network/FLEXNetworkRecorder.m
+++ b/Classes/Network/FLEXNetworkRecorder.m
@@ -177,8 +177,6 @@ NSString *const kFLEXNetworkRecorderResponseCacheLimitDefaultsKey = @"com.flex.r
             return;
         }
     }
-    
-    FLEXHTTPTransaction *transaction = [FLEXHTTPTransaction request:request identifier:requestID];
 
     // Before async block to keep times accurate
     if (redirectResponse) {
@@ -187,6 +185,11 @@ NSString *const kFLEXNetworkRecorderResponseCacheLimitDefaultsKey = @"com.flex.r
     }
 
     dispatch_async(self.queue, ^{
+        FLEXHTTPTransaction *transaction = [FLEXHTTPTransaction request:request identifier:requestID];
+        if (!transaction) {
+            return;
+        }
+
         [self.orderedHTTPTransactions insertObject:transaction atIndex:0];
         self.requestIDsToTransactions[requestID] = transaction;
 
--- a/Classes/Network/FLEXNetworkRecorder.m
+++ b/Classes/Network/FLEXNetworkRecorder.m
@@ -145,7 +145,7 @@ NSString *const kFLEXNetworkRecorderResponseCacheLimitDefaultsKey = @"com.flex.r
         }
 
         [self notify:kFLEXNetworkRecorderTransactionsClearedNotification transaction:nil];
-    }
+    });
 }
 
 - (void)clearExcludedTransactions {
--- a/Classes/Network/FLEXNetworkRecorder.m
+++ b/Classes/Network/FLEXNetworkRecorder.m
@@ -112,38 +112,40 @@ NSString *const kFLEXNetworkRecorderResponseCacheLimitDefaultsKey = @"com.flex.r
 }
 
 - (void)clearRecordedActivity:(FLEXNetworkTransactionKind)kind matching:(NSString *)query {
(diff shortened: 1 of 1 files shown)


NSExceptional:
It won't be nil, look at the code in the initializers; it never returns nil

Those checks are not coming from the initializer, they're because we're pulling a transaction out of a dictionary :)

matrush:
The crash log shows it's `nil`, it's happening :) Both me and #574 got the same issue so believe it.
```
Thread #42 Crashed:
0 libobjc.A.dylib 0x1992333f0 objc_retain + 0x10
1 CoreFoundation 0x1809a8a6c cow_copy_storage + 0x84
2 CoreFoundation 0x1809fcbe0 _cow_mutate_slow + 0x1c
3 CoreFoundation 0x180969710 -[__NSArrayM insertObject:atIndex:] + 0xb0
4 FLEXFramework 0x10ffaca18 __85-[FLEXNetworkRecorder recordRequestWillBeSentWithRequestID:request:redirectResponse:]_block_invoke (FLEXNetworkRecorder.m:190)
5 libdispatch.dylib 0x180665920 _dispatch_call_block_and_release + 0x1c
6 libdispatch.dylib 0x18066766c _dispatch_client_callout + 0x10
7 libdispatch.dylib 0x18066edf0 _dispatch_lane_serial_drain + 0x29c
8 libdispatch.dylib 0x18066f964 _dispatch_lane_invoke + 0x184
9 libdispatch.dylib 0x18067a1b4 _dispatch_workloop_worker_thread + 0x28c
10 libsystem_pthread.dylib 0x1f1d770f0 _pthread_wqthread + 0x11c
```

matrush:
Or probably the problem is `orderedHTTPTransactions` has a race condition? That might make sense as well.

NSExceptional:
I think it was that race condition, there's no way nil gets returned from that initializer unless the app runs out of memory entirely

matrush:
@NSExceptional Sure, I'm not able to repro but I will take another pass to see if I can fix the race.

NSExceptional:
Didn't you just fix it? In the PR I just merged? 😳

NSExceptional:
Oh, that's different isn't it… hang on

NSExceptional:
Well, the `transaction` is not `nil` because `NSMutableArray` simply throws an exception of it is `nil`. Something else is going on.

I don't _think_ there's a race condition with `orderedHTTPTransactions` because we only access it from within our serial queue (right?) but I don't know for sure

NSExceptional:
In the `clearRecordedActivity` method we are not accessing it from the serial queue, so there is a potential race condition. Were you using the "clear" button when the crash occurred? That would explain it

matrush:
I fixed one instance. However that seems to be introduced in https://github.com/FLEXTool/FLEX/commit/02730c6c86fd45a71552cd284045cb65510d77e5, which is ~3 months old. However #574 reported the same crash trace back in 11/22/21. So something else must be going on.

NSExceptional:
Thanks! Leaving the linked issue open because I'm not _sure_ this will resolve it

PR #547, opened 2021-08-01
Update project.pbxproj to fix crash on launch.

duraidabdul:
This reverts some changes in fix #514 that breaks FLEX when it is built with Xcode 12 and 13. Since this change I was getting when running an app linked to FLEX.framework.
```
dyld[4587]: Library not loaded: /Library/Frameworks/FLEX.framework/FLEX
  Referenced from: /private/var/containers/Bundle/Application/redacted/redacted.app/redacted
  Reason: tried: '/Library/Frameworks/FLEX.framework/FLEX' (no such file), '/System/Library/Frameworks/FLEX.framework/FLEX' (no such file)
Library not loaded: /Library/Frameworks/FLEX.framework/FLEX
  Referenced from: /private/var/containers/Bundle/Application//redacted/redacted.app/redacted
  Reason: tried: '/Library/Frameworks/FLEX.framework/FLEX' (no such file), '/System/Library/Frameworks/FLEX.framework/FLEX' (no such file)
(lldb) 
```

PR #597, opened 2022-04-15
Fix compile errors and warnings

canius:
- Fix compile error: add FLEXFirebaseTransaction.mm into project which was broken in commit: d2e0db87731fd16c90462625a8fdfad642611b00
- Fix the root cause for the PR: https://github.com/FLEXTool/FLEX/pull/547
DYLIB_INSTALL_NAME_BASE should keep @rpath when INSTALL_PATH=$(LOCAL_LIBRARY_DIR)/Frameworks
- Fix #import <> error using Carthage, "" instead.
- Fix some warnings caused by headers visibility.
- SPM, Carthage and Cocoapods test passed.

### DIFF
--- a/FLEX.xcodeproj/project.pbxproj
+++ b/FLEX.xcodeproj/project.pbxproj
@@ -139,6 +139,7 @@
 		94AAF0381BAF2E1F00DE8760 /* FLEXKeyboardHelpViewController.h in Headers */ = {isa = PBXBuildFile; fileRef = 94AAF0361BAF2E1F00DE8760 /* FLEXKeyboardHelpViewController.h */; settings = {ATTRIBUTES = (Private, ); }; };
 		94AAF0391BAF2E1F00DE8760 /* FLEXKeyboardHelpViewController.m in Sources */ = {isa = PBXBuildFile; fileRef = 94AAF0371BAF2E1F00DE8760 /* FLEXKeyboardHelpViewController.m */; };
 		94AAF03A1BAF2F0300DE8760 /* FLEXKeyboardShortcutManager.h in Headers */ = {isa = PBXBuildFile; fileRef = 942DCD821BAE0AD300DB5DC2 /* FLEXKeyboardShortcutManager.h */; settings = {ATTRIBUTES = (Private, ); }; };
+		A42069CF2809377F00EBCAEA /* FLEXFirebaseTransaction.mm in Sources */ = {isa = PBXBuildFile; fileRef = A42069CE2809377F00EBCAEA /* FLEXFirebaseTransaction.mm */; };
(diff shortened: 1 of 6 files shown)


NSExceptional:
You tested SPM, Carthage, and Cocoapods in a separate project and they all work when pointed to your PR branch?

FLEX.xcodeproj/project.pbxproj
```diff
@@ -2126,9 +2130,9 @@
 					"$(PROJECT_DIR)",
 				);
 				INFOPLIST_FILE = Classes/Info.plist;
-				INSTALL_PATH = "@rpath";
+				INSTALL_PATH = "$(LOCAL_LIBRARY_DIR)/Frameworks";
 				IPHONEOS_DEPLOYMENT_TARGET = 9.0;
-				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/Frameworks @loader_path/Frameworks";
+				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/../Frameworks @loader_path/Frameworks";
```
Qusic, line 2135: I am not sure about this change.

a typical ios app bundle should look like:
```
MyApp.app
├── MyApp
└── Frameworks
    └── FLEX.framework
        ├── FLEX
        └── Info.plist
```
`@executable_path` should refer to `MyApp.app/MyApp`, then you shouldn't need `../` here.
instead, macOS app bundle put main executable in `MyApp.app/Content/MacOS`, then `../` can make sense:
```
MyApp.app
└── Contents
    ├── Frameworks
    │   └── FLEX.framework
    │       ├── FLEX
    │       └── Info.plist
    └── MacOS
        └── MyApp
```

canius: You're correct. 
The root cause is that DYLIB_INSTALL_NAME_BASE should keep @rpath. 
LD_RUNPATH_SEARCH_PATHS rollbacked.

canius:
> You tested SPM, Carthage, and Cocoapods in a separate project and they all work when pointed to your PR branch?

Yes. Test passed using SPM, Carthage and Cocoapods in a separate project on both ios device and simulator.
 SPM and Carthage pointed to my branch and Cocoapods pointed to my repo using local pod.

NSExceptional:
Awesome. What did you need to change `rpath` for?

canius:
> Awesome. What did you need to change `rpath` for?

The `INSTALL_PATH=rpath` currently is not the standard path for ios framework project. It's confused a lot.
Also this will result in some failures when generating the xcframework using some simple scripts. Have to do some extra works to resolve this.

NSExceptional:
I currently don't generate xcframeworks so that's fine with me I think, as long as it works with Cocoapods and SPM!

NSExceptional:
Thanks :)

PR #600, opened 2022-05-06
Fix implicit casting warnings

matrush:
These are found by `-Wconversion`

- Use explicit cast to avoid implicit cast warnings
- Move params to local variable in blocks instead of using __block variables.

### DIFF
--- a/Classes/Editing/ArgumentInputViews/FLEXArgumentInputTextView.m
+++ b/Classes/Editing/ArgumentInputViews/FLEXArgumentInputTextView.m
@@ -128,7 +128,7 @@
 }
 
 - (CGFloat)inputTextViewHeight {
-    return ceil([[self class] inputFont].lineHeight * self.numberOfInputLines) + 16.0;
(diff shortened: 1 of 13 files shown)


NSExceptional:
What's wrong with __block params?

matrush:
> 

Nothing wrong, but the main difference is scope is different, generally we should define variable at the minimum scope. Here within the block has the minimal scope. You are not using the variables outside of the blocks so I don't see why do we need to define them outside.

NSExceptional:
Once again, I am unable to reproduce this warning; latest version of Xcode, with warnings as errors and this warning enable din the project. I would really like to be able to reproduce these. If you can't help me reproduce them then I would rather recommend you disable that warning for FLEX, or just silence warnings from FLEX entirely if possible.

How does your server compile FLEX? Are you using clang or another compiler?

NSExceptional:
The reason I'm against merging these until I can reproduce them is that without being able to reproduce them, I'm just going to continue introducing these warnings, and you're going to keep making PRs to fix them, and I would rather circumvent that entirely by being able to see them for myself in the first place

PR #601, opened 2022-05-11
allow copy keychain password data

profiles:


profiles:
> You know debug description doesn't necessarily contain the full data right?

I don't know, but it's simple and useful for nsdata

profiles:
> Is this really useful to you?
> 
> image

this is description, not debugDescription

profiles:
> Is this really useful to you?
> 
> image

switch tab to NSObject, and filter "debugDescription", you can see it

NSExceptional:
Ah. I'm still not sure coping the plain string representation is the best option. What if you could airdrop the data as a file or copy a base64 encoded string?

profiles:
> Ah. I'm still not sure coping the plain string representation is the best option. What if you could airdrop the data as a file or copy a base64 encoded string?

it's not the best option, but it's option which can fix a problem when the keychain NSData forced convert to NSString.
base64 string can reduce 2/3 bytes, but hex string is more explicitly.
as for airdrop, it's complex :(

NSExceptional:
Do you mind me asking what you do with the data once you copy the hex string?

profiles:
> Do you mind me asking what you do with the data once you copy the hex string?

view, check, export or analysis keychain data

PR #602, opened 2022-05-11
Fix -Wobjc-signed-char-bool-implicit-int-conversion for FLEXSwiftInternal

matrush:
Fix `-Wobjc-signed-char-bool-implicit-int-conversion` for `FLEXSwiftInternal`

Fixing this trace. If you have the `-Wobjc-signed-char-bool-implicit-int-conversion` enabled and `-Werror` enabled this will fail to build.
```
[CONTEXT] 
stderr: FLEX/Classes/Utility/Runtime/Objc/FLEXSwiftInternal.mm:88:9: error: implicit conversion from integral type 'unsigned long' to 'BOOL' [-Werror,-Wobjc-signed-char-bool-implicit-int-conversion]
[CONTEXT]         return rodata & FAST_IS_SWIFT_STABLE;
[CONTEXT]         ^
[CONTEXT]                (                            ) ? YES : NO
```

See more context here: https://stackoverflow.com/questions/65302104/how-to-enable-the-compiler-option-wobjc-signed-char-bool-implicit-int-conversio

### DIFF
--- a/Classes/Utility/Runtime/Objc/FLEXSwiftInternal.mm
+++ b/Classes/Utility/Runtime/Objc/FLEXSwiftInternal.mm
@@ -85,8 +85,8 @@ extern "C" BOOL FLEXIsSwiftObjectOrClass(id objOrClass) {
     class_data_bits_t rodata = ((__bridge objc_class_ *)(cls))->bits;
     
     if (@available(macOS 10.14.4, iOS 12.2, tvOS 12.2, watchOS 5.2, *)) {
-        return rodata & FAST_IS_SWIFT_STABLE != 0;
+        return (rodata & FAST_IS_SWIFT_STABLE) != 0;
     } else {
-        return rodata & FAST_IS_SWIFT_LEGACY != 0;
+        return (rodata & FAST_IS_SWIFT_LEGACY) != 0;
     }
 }
--- a/Classes/Utility/Runtime/Objc/FLEXSwiftInternal.mm
+++ b/Classes/Utility/Runtime/Objc/FLEXSwiftInternal.mm
@@ -85,8 +85,8 @@ extern "C" BOOL FLEXIsSwiftObjectOrClass(id objOrClass) {
     class_data_bits_t rodata = ((__bridge objc_class_ *)(cls))->bits;
     
     if (@available(macOS 10.14.4, iOS 12.2, tvOS 12.2, watchOS 5.2, *)) {
-        return rodata & FAST_IS_SWIFT_STABLE;
+        return rodata & FAST_IS_SWIFT_STABLE != 0;
     } else {
-        return rodata & FAST_IS_SWIFT_LEGACY;
+        return rodata & FAST_IS_SWIFT_LEGACY != 0;
     }
 }


NSExceptional:
Once again, I am unable to reproduce this warning; latest version of Xcode, with warnings as errors and this warning enable din the project. I would really like to be able to reproduce these. If you're unable to help me reproduce them in Xcode, then I would rather recommend you disable that warning for FLEX, or just silence warnings from FLEX entirely if possible.

How does your server compile FLEX? Are you using clang or another compiler?

The reason I'm against merging these until I can reproduce them is that without being able to reproduce them, I'm just going to continue introducing these warnings, and you're going to keep making PRs to fix them, and I would rather circumvent that entirely by being able to see them for myself in the first place

matrush:
> Still, I am unable to reproduce this warning; latest version of Xcode, with warnings as errors and this warning enabled in the project. If you're unable to help me reproduce them in Xcode, then I would rather recommend you disable that warning for FLEX, or just silence warnings from FLEX entirely if possible :/
> 
> How does your server compile FLEX? Are you using clang or another compiler?
> 
> The reason I'm against merging these until I can reproduce them is that without being able to reproduce them, I'm just going to continue introducing these warnings, and you're going to keep making PRs to fix them, and I would rather circumvent that entirely by being able to see them for myself in the first place

I have the link attached which shows the instructions. I cannot disclose too much detail other than the flags I provided. I don't have the full context but the error was coming from some catalyst builds which is aligned with what the link says.

matrush:
See here: https://developer.apple.com/documentation/apple-silicon/addressing-architectural-differences-in-your-macos-code

> Treat BOOL Variables as Binary Values
> As a rule, the Objective-C [BOOL](https://developer.apple.com/documentation/objectivec/bool) type has only two appropriate values: YES or NO. On Apple silicon, the compiler defines the BOOL type to be a native bool, but on Intel-based Mac computers, it is a signed char. To avoid issues in universal binaries:
> 
> Never perform mathematical operations on [BOOL](https://developer.apple.com/documentation/objectivec/bool) variables.
> 
> Never increment or decrement BOOL variables.
> 
> Never assume the numerical value of a BOOL is anything other than 0 or 1.
> 
> To illustrate the problem, consider the following example:
> 
> int nBytes = 1024;
> BOOL receivedBytes = nBytes;
> 
> if (receivedBytes) {
>    printf("Success!\n");
> } else {
>    printf("Failure...\n");
> }
> On the x86_64 architecture, receivedBytes evaluates to 0, or false. On the arm64 architecture, receivedBytes evaluates to true. One way to fix the preceding code is to include two negation operators (!!) in front of the nBytes variable before assigning it to the receivedBytes variable, as shown in the following example:
> 
> int nBytes = 1024;
> BOOL receivedBytes = !!nBytes;
> Note
> 
> Enable the -Wobjc-signed-char-bool-implicit-int-conversion compiler option to generate warnings for implicit [BOOL](https://developer.apple.com/documentation/objectivec/bool) conversions at compile time.

NSExceptional:
Instead of linking me the instructions that tell me to do what I've already tried, can you try it yourself in an Xcode project so you can see I'm not taking crazy pills? haha

NSExceptional:
Mac Catalyst you say? That's interesting

While FLEX doesn't claim to support Catalyst, I'm glad to know it at least compiles! I just tried to build it myself and there's a bunch of deprecation warnings. I still don't see this warning, though

matrush:
Yeah, for some reason I didn't get the warning in Xcode too with the setting. Maybe there is another magic setting somewhere.

However I think this is almost harmless to merge, I think having a repro is better but this change is technically reasonable so I'd vote for fixing. I don't mind help fixing it moving forward if circumvent is not possible. As long as you are building for the major flow I think is fine.

NSExceptional:
I really don't want to adjust my code style for something that doesn't generate a warning seemingly at all except in this outlier case; can you add a flag to ignore this warning in your CI build config? `-Wno-<the-warning>` should do it

This is supposed to be allowed, too, I think; Apple has encouraged implicit Boolean conversion code for a long time in their documentation. I'm quite fond of it tbh

NSExceptional:
I really like the clarifying parenthesis commit though, happy to merge that right away

matrush:
It's not about coding style, it's technically different on Apple silicon as Apple pointed out in their article. I find it makes no sense that you are asking people to turn off warning instead of letting a proper fix going through. This is wasting contributor's time and your time as well.

NSExceptional:
Yes, the PR is perfectly benign. The code change itself is not the issue. It's the fact that you're _constantly_ making PRs to fix things I can't reproduce, and if I can't reproduce them there's nothing stopping me from introducing more or even re-introducing ones that have already been fixed. This isn't anyone's fault, and I'm certainly not saying there's anything wrong with you going out of your way to help me fix these warnings I can't see, but it still bothers me that I can't see them.

I'm asking for your help here. I will merge this, but before you make another PR to fix a warning that doesn't appear in vanilla Xcode, I would greatly appreciate some research on your part as to why that error appears elsewhere and not in vanilla Xcode. I want to fix the _source_ of these issues, not continue to put bandaids on it. I don't think I'm being unreasonable, but if you really can't help any further, then maybe it is best for you to save yourself from future hassle and just silence the "invisible" warnings for the FLEX target in your CI?

NSExceptional:
I asked my peers about this and we were able to reproduce it; I guess neither of us tried very hard to reproduce it because apparently the _"Implicit Boolean Conversions"_ warning is _not_ the same thing as this flag; adding it by hand makes the warning appear 🤡 

I was also able to write this test in a blank project to see the dangers of doing this myself:
screen_shot_2022-05-12_at_12 49 36_pm

Apparently boolean conversion is only okay when the expression is being evaluated by `if/while` conditions or being passed in as a BOOL argument. It's really scary how inconsistent the behavior is here. I guess the compiler will warn us about it from now on if I add this warning to the project.

If you make another PR like this, can you add the warning flag to the project to make future warnings appear? 🙏🏻

matrush:
I tried adding the flag as followed by the link I mentioned earlier but it's compiling file for me somehow. Are you able to screenshot/record or share the thing that shows warning on your end?

NSExceptional:
Screen Shot 2022-05-12 at 1 59 23 PM
Screen Shot 2022-05-12 at 1 59 38 PM

matrush:
Weird, that's exact my setup. Maybe I can try again once you push this to master.

PR #609, opened 2022-07-14
Fixed issue where websocket hook did not call original completion

ogres:
Send message hook was changing original completion instead of wrapping it in a hook.
Similar hook already exists for receivemessage functionality, but it was not working for send-message, original completion was never called.
Adjusted code to create a new completionHook which records the information and then original completion block.

I also updated API-Key for piesocket and added a message after sending message, to make sure that completion is being called and at least success/failure message is logged.

### DIFF
--- a/Classes/Network/PonyDebugger/FLEXNetworkObserver.m
+++ b/Classes/Network/PonyDebugger/FLEXNetworkObserver.m
@@ -1591,15 +1591,19 @@ static FIRDocumentReference * _logos_method$_ungrouped$FIRCollectionReference$ad
         [FLEXNetworkObserver.sharedObserver
             websocketTask:slf sendMessagage:message
         ];
-        completion = ^(NSError *error) {
+        
+        id completionHook = ^(NSError *error) {
             [FLEXNetworkObserver.sharedObserver
                 websocketTaskMessageSendCompletion:message
                 error:error
             ];
+            if (completion) {
+                completion(error);
+            }
         };
         
         ((void(*)(id, SEL, id, id))objc_msgSend)(
-            slf, swizzledSelector, message, completion
+            slf, swizzledSelector, message, completionHook
         );
     };
 
--- a/Example/FLEXample/MiscNetworkRequests.m
+++ b/Example/FLEXample/MiscNetworkRequests.m
@@ -153,7 +153,7 @@
 }
 
 - (void)sendExampleWebsocketTraffic:(NSURLSession *)sessionWithDelegate {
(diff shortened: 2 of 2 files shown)


ogres:
@NSExceptional could you have a look at this one?

Example/FLEXample/MiscNetworkRequests.m
```diff
@@ -153,7 +153,7 @@ - (void)sendExampleNetworkRequests:(NSURLSession *)sessionWithDelegate {
 }
 
 - (void)sendExampleWebsocketTraffic:(NSURLSession *)sessionWithDelegate {
-    NSString *APIKey = @"oCdCMcMPQpbvNjUIzqtvF1d2X2okWpDQj4AwARJuAgtjhzKxVEjQU6IdCjwm";
+    NSString *APIKey = @"VCXCEuvhGcBDP7XhiJJUDvR1e1D3eiVjgZ9VRiaV";
```
NSExceptional, line 156: Why did you change the API key? Did the old one expire or something?

Classes/Network/PonyDebugger/FLEXNetworkObserver.m
```diff
@@ -1591,15 +1591,17 @@ typedef void (^SendMessageBlock)(
...
             websocketTask:slf sendMessagage:message
         ];
-        completion = ^(NSError *error) {
+        
+        id completionHook = ^(NSError *error) {
             [FLEXNetworkObserver.sharedObserver
                 websocketTaskMessageSendCompletion:message
                 error:error
             ];
+            completion(error);
```
NSExceptional, line 1600: `completion` is nullable, this needs to check for that before calling it or it will crash

NSExceptional (changes requested):
Nice catch, thank you! Also, sorry, I must have totally missed this PR. Thanks for pinging me.

Classes/Network/PonyDebugger/FLEXNetworkObserver.m
NSExceptional, line 1600:
```suggestion
            if (completion) {
                completion(error);
            }
```

Example/FLEXample/MiscNetworkRequests.m
ogres: Yes, it returns error `{"error":"Invalid API key"}`.
Based on piesocket's website, its changed every week:

image

ogres:
> Nice catch, thank you! Also, sorry, I must have totally missed this PR. Thanks for pinging me.

I've added nullability check
Thanks for the review 👍

PR #610, opened 2022-07-17
Xcode 14 - Warnings related to iOS 11 support (Swift Package Manager)

kikeenrique:
Hi,
I would like to contribute fixing a recent warning.

The new Xcode 14 uses iOS SDK 11.0 as base, and compiling in this Xcode shows a warning.
Handle this restriction relying on swift version.

Regards and comments welcomed

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -9,7 +9,7 @@ enum FLEXBuildOptions {
 
 #if swift(>=5.7)
 let platforms: [PackageDescription.SupportedPlatform] = [.iOS(.v11)]
-#elseif swift(>=5.0)
+#else
 let platforms: [PackageDescription.SupportedPlatform] = [.iOS(.v10)]
 #endif
 
--- a/Package.swift
+++ b/Package.swift
@@ -7,9 +7,15 @@ enum FLEXBuildOptions {
     static let silenceWarnings = false
 }
 
+#if swift(>=5.7)
+let platforms: [PackageDescription.SupportedPlatform] = [.iOS(.v11)]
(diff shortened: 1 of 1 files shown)


NSExceptional:
Hi, thank you!

How does this compile though? Shouldn't there be a plain `#else` instead of an `#elseif`?

kikeenrique:
I don't think it's needed to provide all _if_ alternatives to be compilable.
In this case, 5.0 is prior to package file swift-tools-version 5.3, which correspond to swift version 5.3. So it won't be possible to compile with a version out of the covered toolchain.

I did tested it in a custom project, I will test it on the example SPM in the repo, just in case.

kikeenrique:
An unexpected problem, example SPM project uses Reflex (I just learned this existed) instead of Flex package.
Also, current example SPM project  doesn't compile, as it points to a local directory 
`repositoryURL = "file:///Users/tanner/Repos/Reflex";`

To check that it was working, in my fork I've updated example SPM project in branch _update-example_ to use a local Flex package, which uses SPM from the local repo (it uses a relative path), instead of resolving to GitHub

NSExceptional:
I guess it's not technically required since it will compile as long as you're using swift 5 at least, but for clarity's sake, can we change the else to just an `#else` and remove the condition?

kikeenrique:
Yes, of course. Tested and pushed.

kikeenrique:
Hi, I've updated branch with latests commits. It's still in good shape.

NSExceptional:
I missed your reply, whoops! Thank you! Merging now

PR #618, opened 2022-08-30
Use queue instead of self.queue in FLEXNetworkRecorder.m

matrush:
Small bug introduced in https://github.com/FLEXTool/FLEX/commit/b45c655ba2432f827d6b07faf9844a9111c51716


NSExceptional:
Thanks :D good catch

PR #619, opened 2022-08-30
Allow unloading of FLEX window and view controller

inzinga:
FLEX manager singleton stays, once created.  Use `[[FLEXManager sharedManager] unloadExplorer];` to hide and unload FLEX view controller and window.

Unloading these items unhooks all swizzled methods.  They'll be recreated next time `[[FLEXManager sharedManager] showExplorer];` is called.

### DIFF
--- a/Classes/Manager/FLEXManager.h
+++ b/Classes/Manager/FLEXManager.h
@@ -20,6 +20,7 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)showExplorer;
 - (void)hideExplorer;
 - (void)toggleExplorer;
+- (void)unloadExplorer;
 
 /// Programmatically dismiss anything presented by FLEX, leaving only the toolbar visible.
 - (void)dismissAnyPresentedTools:(void (^_Nullable)(void))completion;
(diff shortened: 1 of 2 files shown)


NSExceptional:
This does not unhook swizzled methods.

What's the use case for that anyway, if you don't mind me asking?

inzinga:
Want to integrate this into large product, in debug configurations, of course.  Others are concerned that while it is inactive (`... hideExplorer`) that there is the potential it has some background influence for other parts of our product.

Concerned about `FlexWindow.m `+ (void)initialize`, plus other locations in the package where swizzling is used.

Basically, when we are not using (after having used it), we want to get it completely out of the debug environment.  That is what I was trying to achieve here with exception of the `FlexManager` singleton.

Please advise.

NSExceptional:
Un-swizzling methods takes a lot of work. You basically have to swizzle it in reverse. It would duplicate a lot of code.

Is there any reason you need to have FLEX at all times? Can you create a second debug build target  that doesn't include FLEX?

inzinga:
We'll create another debug target, as you suggest.  Closing this PR.  Thanks for your help, Tanner.  Team here the features offered in FLEX.

PR #616, opened 2022-08-24
Fix Swift Package Manager integration

pyby:
See #615 issue.

PR #620, opened 2022-08-31
Fix missing linked file in Headers directory

BandarHL:
When you install FLEX from Swift Package Manager you get this error, because ```NSDateFormatter+FLEX.h``` it's missing from Headers directory
Screen Shot 2022-08-31 at 6 27 37 PM


NSExceptional:
#616 addresses this; I need to try and get that mergeable today.

Issue #621, opened 2022-08-31
Crash when opening Realm DB table on debug


nathandud:
### Environment
- Platform+version: **iOS 15.5** 
- FLEX version: **7.4.0** 

### Bug Report

Opening a Realm database on a debug build causes a crash due to an `NSParameterAssert` failure in `FLEXTableContentViewController` seen [here](https://github.com/FLEXTool/FLEX/blob/fb2a33876e33a605f59290c135f2017d7ce73199/Classes/GlobalStateExplorers/DatabaseBrowser/FLEXTableContentViewController.m#L59) and copied below.

```
- (instancetype)initWithColumns:(NSArray<NSString *> *)columnNames
                           rows:(NSArray<NSArray<NSString *> *> *)rowData
                         rowIDs:(nullable NSArray<NSString *> *)rowIDs
                      tableName:(nullable NSString *)tableName
                       database:(nullable id<FLEXDatabaseManager>)databaseManager {
    // Must supply all optional parameters as one, or none
    BOOL all = rowIDs && tableName && databaseManager;
    BOOL none = !rowIDs && !tableName && !databaseManager;
    NSParameterAssert(all || none);

    //....
}
```

When opening up a SQLLite database table, the `databaseManager` points to null (and so do `rowIDs` and `tableName` and the `none` condition is satisfied. However when trying to open a Realm database table, the `databaseManager` object pointer is not null and the assert fails. When I remove that assert, the code works as expected and I can view a Realm database.

PR #622, opened 2022-08-31
Remove assert that blocks Realm

nathandud:
Proposed fix to #621. A realm database table opens as expected when this `NSParameterAssert` is removed.

### DIFF
--- a/Classes/GlobalStateExplorers/DatabaseBrowser/FLEXTableContentViewController.m
+++ b/Classes/GlobalStateExplorers/DatabaseBrowser/FLEXTableContentViewController.m
@@ -53,11 +53,6 @@
                          rowIDs:(nullable NSArray<NSString *> *)rowIDs
                       tableName:(nullable NSString *)tableName
                        database:(nullable id<FLEXDatabaseManager>)databaseManager {
-    // Must supply all optional parameters as one, or none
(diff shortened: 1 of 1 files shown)


NSExceptional:
Can you see if just removing the ` && !databaseManager` from the `none` condition part fixes it? I still want to keep the assertion, it just needs to be correct

lubelio:
rowIDs:(nullable NSArray<NSString *> *)rowIDs

> Can you see if just removing the ` && !databaseManager` from the `none` condition part fixes it? I still want to keep the assertion, it just needs to be correct

It only works if removing rowIDs and !rowIDs from both `all` and `none` conditions

NSExceptional:
@lubelio the original issue states:

> When opening up a SQLLite database table, the `databaseManager` points to null (and so do `rowIDs` and `tableName`) and the none condition is satisfied. However when trying to open a Realm database table, the `databaseManager` object pointer is not null and the assert fails. When I remove that assert, the code works as expected and I can view a Realm database.

If this is accurate, then simply removing the `databaseManager` from each check would suffice. The resulting condition would be:

```objc
!rowIDs && !tableName &&
```

And this would satisfy Realm, since those are both still `nil`. Right?

PR #626, opened 2022-09-13
Podspec: update xcconfig to pod_target_xcconfig

JerryZQS:
Specify xcconfig settings in spec only for pod, not for consuming project.


NSExceptional:
Didn't know that was a thing! Thanks!

PR #629, opened 2022-09-23
ObjectExplorer: Add a new Shortcut for UIWindow to change animations speed

AnthoPakPak:
This PR adds a new Shortcut to UIWindow to change its animations speed. It basically replicates the "Slow Animations" feature, only available on the Simulator. The user can set the speed of his choice.

As explained in the comments, as of now I've added it to UIView Shortcuts class instead of creating a new one for UIWindow. This is done like this to prevent UIWindow Shortcuts to override UIView ones (since as of now, an object can only get the Shortcuts from one class). Having a separate UIWindow class would have imply duplicating current UIView Shortcuts, which is not desirable. Tell me if you'd prefer any other way of handling this situation.

### DIFF
--- a/Classes/ObjectExplorers/Sections/Shortcuts/FLEXWindowShortcuts.m
+++ b/Classes/ObjectExplorers/Sections/Shortcuts/FLEXWindowShortcuts.m
@@ -1,0 +1,45 @@
+//
+//  FLEXWindowShortcuts.m
+//  FLEX
+//
+//  Created by AnthoPak on 26/09/2022.
+//
+
+#import "FLEXWindowShortcuts.h"
+#import "FLEXShortcut.h"
+#import "FLEXAlert.h"
+#import "FLEXObjectExplorerViewController.h"
+
+@implementation FLEXWindowShortcuts
+
+#pragma mark - Overrides
+
++ (instancetype)forObject:(UIView *)view {
+    return [self forObject:view additionalRows:@[
+        [FLEXActionShortcut title:@"Animations Speed" subtitle:^NSString *(UIWindow *window) {
+            return [NSString stringWithFormat:@"Current speed: %.2f", window.layer.speed];
+        } selectionHandler:^(UIViewController *host, UIWindow *window) {
+            [FLEXAlert makeAlert:^(FLEXAlert *make) {
+                make.title(@"Change Animations Speed");
+                make.message([NSString stringWithFormat:@"Current speed: %.2f", window.layer.speed]);
+                make.configuredTextField(^(UITextField * _Nonnull textField) {
+                    textField.placeholder = @"Speed value";
+                    textField.keyboardType = UIKeyboardTypeDecimalPad;
+                });
+                
+                make.button(@"OK").handler(^(NSArray<NSString *> *strings) {
+                    CGFloat speedValue = strings.firstObject.floatValue;
+                    window.layer.speed = speedValue;
+
+                    // Refresh the host view controller to update the shortcut subtitle, reflecting current speed
+                    [(FLEXObjectExplorerViewController *)host reloadData];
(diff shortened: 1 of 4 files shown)


NSExceptional:
Can we make this just a shortcut on UIWindow for the time being without modifying the view shortcuts file? One of them doesn't even work for UIWindow so we don't lose that much, right?

AnthoPakPak:
You're right, I did not noticed "Nearest VC" wasn't working on UIWindows.
Just made the changes to move this shortcut where it truly belongs.

AnthoPakPak:
Ugh, just clicked the wrong button while sending my comment, which closed the PR. Reopening.

Classes/ObjectExplorers/Sections/Shortcuts/FLEXWindowShortcuts.m
```diff
@@ -0,0 +1,45 @@
...
+#import "FLEXAlert.h"
+#import "FLEXObjectExplorerViewController.h"
+
+@implementation FLEXWindowShortcuts
+
+#pragma mark - Overrides
+
++ (instancetype)forObject:(UIView *)view {
+    return [self forObject:view additionalRows:@[
+        [FLEXActionShortcut title:@"Animations Speed" subtitle:^NSString *(UIWindow *window) {
```
NSExceptional, line 19:
```suggestion
        [FLEXActionShortcut title:@"Animation Speed" subtitle:^NSString *(UIWindow *window) {
```

Classes/ObjectExplorers/Sections/Shortcuts/FLEXWindowShortcuts.m
```diff
@@ -0,0 +1,45 @@
...
+            return [NSString stringWithFormat:@"Current speed: %.2f", window.layer.speed];
+        } selectionHandler:^(UIViewController *host, UIWindow *window) {
+            [FLEXAlert makeAlert:^(FLEXAlert *make) {
+                make.title(@"Change Animations Speed");
```
NSExceptional, line 23:
```suggestion
                make.title(@"Change Animation Speed");
```

Classes/ObjectExplorers/Sections/Shortcuts/FLEXWindowShortcuts.m
```diff
@@ -0,0 +1,45 @@
...
+                make.message([NSString stringWithFormat:@"Current speed: %.2f", window.layer.speed]);
+                make.configuredTextField(^(UITextField * _Nonnull textField) {
+                    textField.placeholder = @"Speed value";
```
NSExceptional, line 26:
```suggestion
                    textField.placeholder = @"Default: 1.0";
```

Classes/ObjectExplorers/Sections/Shortcuts/FLEXWindowShortcuts.m
```diff
@@ -0,0 +1,45 @@
...
+                    textField.keyboardType = UIKeyboardTypeDecimalPad;
+                });
+                
+                make.button(@"OK").handler(^(NSArray<NSString *> *strings) {
+                    CGFloat speedValue = strings.firstObject.floatValue;
+                    window.layer.speed = speedValue;
+
+                    // Refresh the host view controller to update the shortcut subtitle, reflecting current speed
+                    [(FLEXObjectExplorerViewController *)host reloadData];
```
NSExceptional, line 35: Hmmm I don't like this but I guess that means I need a way to reload the item itself from this callback
```suggestion
                    // TODO: this shouldn't be necessary
                    [(FLEXObjectExplorerViewController *)host reloadData];
```

NSExceptional:
Thanks!

PR #632, opened 2022-10-13
Crash when trying to change windowLevel in iOS16.

wani-prnd:
Crash when trying to change windowLevel in iOS16.

- *** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'The status bar window cannot be accessed from within an application.' terminating with uncaught exception of type NSException


pokerJson:
you can download source code，I don't think the author pushed it。

NSExceptional:
`-statusWindow` already has the appropriate guard statement on `master`. I am publishing an update today though so your cocoapods/SPM version will be updated.

PR #642, opened 2022-12-01
Fix animation speed modification in other locales

Snowy1803:
In some locales where the decimal separator isn't a dot, the animation speed editor doesn't work with decimal value.

The `UIKeyboardTypeDecimalPad` keyboard has a comma key in French, but it is then parsed expecting a dot for the decimal.

I changed it to use `NSNumberFormatter`, which is locale-aware, and correctly expects a comma as a decimal separator in such languages.


NSExceptional:
Awesome! Can you change the second line to use property syntax for the setter? Then I can merge this.

Thank you so much!

Snowy1803:
Can we get this merged?

NSExceptional:
Idk why but the GitHub all has stopped sending me push notifications, I'm sorry!

PR #644, opened 2022-12-12
Fix cURL null payload for compressed HTTP bodies

sergiosette:
Issue:
- When tapping the "Copy curl" button the cURL string has missing request body (it shows as null) when the content is compressed (gzip, deflate)

Fix:
- I realized that the library already has a way to decode compressed payloads since the request body shows fine in the UI. So I just moved the check for `Content-Encoding` to `FLEXUtility` to reuse in the `FLEXNetworkCurlLogger` and call the inflate method from `FLEXUtility` when necessary.


NSExceptional:
Thank you!

PR #645, opened 2022-12-25
Adding fix to ITMS-90338: Non-public API usage. 

joseluissanchezporrogodoy:
To upload app in the app store that contain FLEX

ITMS-90338: Non-public API usage - The app references non-public selectors in IzzigoMirada:  _setWidth:


NSExceptional:
Do not include FLEX in App Store builds

joseluissanchezporrogodoy:
Ok thanks!!! And sorry

PR #649, opened 2023-01-20
Bazel

h0tv1t:
Nice


NSExceptional:
Did you just open a PR with the changes from the Bazel branch? We can't merge that, it breaks everything else in it's current state

PR #651, opened 2023-02-10
Add support for nested global entries

revolter:
Please review each commit separately, as this PR is logically split into multiple "slow" transitions from the current implementation to the new one.

### DIFF
--- a/Classes/GlobalStateExplorers/Globals/FLEXGlobalsEntry.h
+++ b/Classes/GlobalStateExplorers/Globals/FLEXGlobalsEntry.h
@@ -10,6 +10,8 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+@class FLEXUserGlobalEntriesContainer;
+
 typedef NS_ENUM(NSUInteger, FLEXGlobalsRow) {
     FLEXGlobalsRowProcessInfo,
     FLEXGlobalsRowNetworkHistory,
@@ -51,8 +53,10 @@ typedef UIViewController * _Nullable (^FLEXGlobalsEntryViewControllerFuture)(voi
 /// Do something like present an alert, then use the host
 /// view controller to present or push another view controller.
 typedef void (^FLEXGlobalsEntryRowAction)(__kindof UITableViewController * _Nonnull host);
+/// Use the container to register nested global entries.
+typedef void (^FLEXNestedGlobalEntriesHandler)(FLEXUserGlobalEntriesContainer * _Nonnull container);
 
-/// For view controllers to conform to to indicate they support being used
+/// For view controllers to conform and to indicate they support being used
 /// in the globals table view controller. These methods help create concrete entries.
 ///
 /// Previously, the concrete entries relied on "futures" for the view controller and title.
--- a/Classes/GlobalStateExplorers/Globals/FLEXGlobalsEntry.h
+++ b/Classes/GlobalStateExplorers/Globals/FLEXGlobalsEntry.h
@@ -85,15 +85,20 @@ typedef void (^FLEXNestedGlobalEntriesHandler)(FLEXUserGlobalEntriesContainer * 
 @interface FLEXGlobalsEntry : NSObject
 
 @property (nonatomic, readonly, nonnull)  FLEXGlobalsEntryNameFuture entryNameFuture;
+@property (nonatomic, readonly)           UITableViewCellAccessoryType cellAccessoryType;
 @property (nonatomic, readonly, nullable) FLEXGlobalsEntryViewControllerFuture viewControllerFuture;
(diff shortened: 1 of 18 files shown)


revolter:
In Swift, if you don't use the closure argument, like this:
```swift
FLEXManager.shared.registerGlobalEntry(withName: "Name") { _ in
    print("Log")
}
```
you now get the build error:
```
Ambiguous use of 'registerGlobalEntry'
```
and the only workaround that I can think of is changing the call to:
```swift
FLEXManager.shared.registerGlobalEntry(withName: "Name") { (_: UITableViewController) in
    print("Log")
}
```
or
```swift
FLEXManager.shared.registerGlobalEntry(withName: "Name") { (_: FLEXUserGlobalEntriesContainer) in
    print("Log")
}
```
But this means that you have to look into FLEX's files to find this out, if you're new to it.

---

So, I will rename the new method (from Swift's POV, the first parameter name), to avoid this ambiguity:
```swift
FLEXManager.shared.registerNestedGlobalEntry(withName: "Name") { _ in
    print("Log")
}
```

revolter:
@NSExceptional, do you have time to look over this? 😁

NSExceptional:
Not at the moment no; I was actually in a serious car accident on Tuesday and just got out of the hospital yesterday 🤕

I reviewed a PR yesterday but it was just a small bug fix

NSExceptional:
I'll try to get to this when I can use my hand again

revolter:
I also added support for optionally customizing the cell accessory type:

Before | After
--- | ---
 |

NSExceptional:
Hey dawg! I'm finally getting around to clearing out all these PRs and issues that have accumulated over the last year. Sorry for the delay. I appreciate your patience (and frequent contributions).

Still reviewing, but I'm curious, what's the use case for changing the cell accessory type?

revolter:
@NSExceptional, would you have some time to check this, and hopefully merge it? 🙏🏻

It would reduce the overhead of using and syncing a fork in order to use this functionality.

Classes/GlobalStateExplorers/Globals/FLEXGlobalsEntry.h
```diff
@@ -81,15 +85,20 @@ typedef void (^FLEXGlobalsEntryRowAction)(__kindof UITableViewController * _Nonn
...
+             cellAccessoryType:(UITableViewCellAccessoryType)cellAccessoryType
+                           row:(FLEXGlobalsRow)row;
 
 + (instancetype)entryWithNameFuture:(FLEXGlobalsEntryNameFuture)nameFuture
+                  cellAccessoryType:(UITableViewCellAccessoryType)cellAccessoryType
                viewControllerFuture:(FLEXGlobalsEntryViewControllerFuture)viewControllerFuture;
 
 + (instancetype)entryWithNameFuture:(FLEXGlobalsEntryNameFuture)nameFuture
+                  cellAccessoryType:(UITableViewCellAccessoryType)cellAccessoryType
                              action:(FLEXGlobalsEntryRowAction)rowSelectedAction;
```
NSExceptional: In addition to these new ones, can we keep the original signatures around and have them default to the disclosure indicator? Removing them is a breaking change I'd like to avoid

Classes/GlobalStateExplorers/Globals/FLEXGlobalsViewController.m
NSExceptional: I probably need to just set up a pre-commit linter hook, but in the meantime can we discard all the trailing whitespace changes?

Classes/GlobalStateExplorers/Globals/FLEXUserGlobalEntriesContainer.h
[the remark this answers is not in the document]
NSExceptional: Also, can we rename this to `Custom` instead of `User`?

Classes/Manager/FLEXManager+Extensibility.h
```diff
@@ -25,7 +29,7 @@ NS_ASSUME_NONNULL_BEGIN
 /// The objectFutureBlock will be invoked from the main thread and may return nil.
 /// @note The passed block will be copied and retain for the duration of the application,
 /// you may want to use __weak references.
-- (void)registerGlobalEntryWithName:(NSString *)entryName objectFutureBlock:(id (^)(void))objectFutureBlock;
+- (void)registerGlobalEntryWithName:(NSString *)entryName objectFutureBlock:(id (^)(void))objectFutureBlock __attribute((deprecated("Use the same method on the globalEntriesContainer property instead.")));
```
NSExceptional: I see no reason to depreciate the old method, I'm not in a hurry to add more warnings for people to deal with
```suggestion
- (void)registerGlobalEntryWithName:(NSString *)entryName objectFutureBlock:(id (^)(void))objectFutureBlock;
```

PR #659, opened 2023-03-07
[info]fix the name is nil cause crash

ZuhanLin:
fix crash


ZuhanLin:
@NSExceptional  pls check this case!

NSExceptional:
Thank you! Do you have a crash log? I'm curious

ZuhanLin:
image

I got this crash when i click UITableView info button

image

PR #660, opened 2023-03-08
Fix EXC_BAD_ACCESS in FLEXNetworkObserver

ahmedgomaa27:
the crash happened after I used FLEX and Firestore in my app, the root cause for it is trying to set error in Orig block and the
error value is nil so we need to check on it first  
Screen Shot 2023-03-08 at 2 02 53 PM

### DIFF
--- a/Classes/Network/PonyDebugger/FLEXNetworkObserver.m
+++ b/Classes/Network/PonyDebugger/FLEXNetworkObserver.m
@@ -285,7 +285,9 @@ static void _logos_method$_ungrouped$FIRDocumentReference$setData$merge$completi
     void (^orig)(NSError *) = completion;
     completion = ^(NSError *error) {
         [FLEXNetworkRecorder.defaultRecorder recordFIRDidSetData:error transactionID:requestID];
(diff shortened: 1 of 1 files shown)


NSExceptional:
Passing nil to a function doesn't cause that error. I think `orig` may be nil or corrupted instead

ahmedgomaa27:
@NSExceptional  when i run XCode debugging , i found that error = nil

NSExceptional:
Yes but that won't cause a crash, error is supposed to be nil unless there is an error

ryanolsonk:
I've seen this one. The completion is optional so `orig` can (validly) be nil. We should just guard against that.

ahmedgomaa27:
@NSExceptional  @ryanolsonk  I updated the code now please check

ryanolsonk:
Looks good to me, @NSExceptional ?

NSExceptional:
Thanks!!

ahmedgomaa27:
@NSExceptional  will be a new version for this fix? as i need to upgrade current version

PR #670, opened 2023-05-01
Fix a view controllers leaking when dismissed via escape key.

timonus:
I found that some view controllers (for example: FLEXHierarchyViewController) are effectively leaked when they're dismissed via the escape key. It looks like they're added as tabs to `FLEXTabList`, but only manually removed from `FLEXTabList` via their own dismiss callbacks. Given that the escape key shortcut bypasses custom dismiss methods and just calls `-dismissVIewController:animated:`, `FLEXTabList` ends up holding a strong reference to a view controller that's never used again.

### DIFF
--- a/Classes/Manager/FLEXManager+Extensibility.m
+++ b/Classes/Manager/FLEXManager+Extensibility.m
@@ -17,6 +17,7 @@
 #import "FLEXFileBrowserController.h"
 #import "FLEXArgumentInputStructView.h"
 #import "FLEXUtility.h"
+#import "FLEXTabList.h"
 
 @interface FLEXManager (ExtensibilityPrivate)
 @property (nonatomic, readonly) UIViewController *topViewController;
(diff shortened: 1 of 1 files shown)


timonus:
Note: this might not be the best solution. I could alternatively see `FLEXTabList` using a weak collection to keep track of tabs, or moving tab removal to some other place besides custom `Done` button callbacks.

NSExceptional:
The tab list is intended to keep track of open tabs; dismissing a tab with a swipe down gesture (or in this case, the escape key) is expected to keep the tab alive so you can revisit it. There is a UI to present a list of all open tabs, I believe by holding down on the globe icon.

If you think the escape key should instead function as pressing the "Done" button—which closes the tab instead of hiding it—I'm open to hearing why; I haven't given keyboard shortcuts much thought really

timonus:
Hmm, alright. I'll update the leak detector I'm using to ignore FLEX if this is intended.

PR #677, opened 2023-08-24
Ignore AVAssetDownloadURLSession to avoid crashing. (FLEXTool#276)

### LINKED ISSUE #276: FLEXNetworkObserver crashes when used with AVAssetDownloadTask or AVAggregateAssetDownloadTask
If you use a `AVAssetDownloadURLSession` for downloading video assets, this will still be used with the network logging facility in FLEX. However, the tasks used for this `NSURLSession` subclass as instances of `AVAssetDownloadTask` or `AVAggregateAssetDownloadTask`. Both of these classes declare that the properties `originalRequest`, `currentRequest` and `response` are unavailable. However this line in `FLEXNetworkObserver`:

https://github.com/Flipboard/FLEX/blob/master/Classes/Network/PonyDebugger/FLEXNetworkObserver.m#L1115

attempts to access those properties (when network logging in enabled) and thus crashes with this exception:

> *** Terminating app due to uncaught exception 'NSGenericException', reason: 'AVAggregateAssetDownloadTask does not support currentRequest property'

I'm not sure what would be the solution here. Perhaps don't use network logging with instances of `AVAssetDownloadURLSession`?

rnapier:
AVAggregateAssetDownloadTask deeply does not like to be looked at. Accessing -currentRequest or -originalRequest will crash. Do not try to observe these.

Fixes #276

### DIFF
--- a/Classes/Network/PonyDebugger/FLEXNetworkObserver.m
+++ b/Classes/Network/PonyDebugger/FLEXNetworkObserver.m
@@ -23,6 +23,7 @@
 #import "FLEXMethod.h"
 #import "Firestore.h"
 
+#import <AVFoundation/AVFoundation.h>
 #import <objc/runtime.h>
 #import <objc/message.h>
 #import <dispatch/queue.h>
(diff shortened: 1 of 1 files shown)


kikeenrique:
I've been using a code quite similar to this to avoid crashes when using FLEX in an app that downloads video.
Is it possible to be merged? Or is there any problem?

NSExceptional (approved):

PR #680, opened 2023-10-16
Fix: when data is nil, `isValidJSONData` will crash

ZuhanLin:
fix crash


NSExceptional:
I would rather assert data is not nil with `NSParameterAssert` and do whatever we need to do first to not even present the view controller if there is no data

PR #683, opened 2023-11-13
Warning removed

superHELLd:


NSExceptional:
I'm not sure what the point of this PR was, but I'm not removing all that code without a more thorough explanation

PR #687, opened 2024-01-30
Use requiredProperties and optionalProperties in debug description

ingridwang:
On iOS 10 and above, `requiredProperties` and `optionalProperties` are preferred over the deprecated `properties`.

Since FLEX only requires iOS 9 and above, a fallback is provided for older versions.


NSExceptional (approved):

NSExceptional:
Thanks!

PR #688, opened 2024-02-03
Fix Xcode 15 SPM warning related to iOS 12 support

kikeenrique:
This PR fixes a Xcode 15 SPM warning related to iOS 12 support, which is the minimum version supported.


NSExceptional (approved):

NSExceptional:
Thanks!

Issue #691, opened 2024-02-10
[CRASH] Thread 1: EXC_BAD_ACCESS (code=1, address=0x10) (FLEXNetworkObserver.m:288)


programming086:
### Environment
- Platform+version: **iOS 15+**
- FLEX version: **5.22.10**

- Firebase
* Firebase 10.21.0
* FirebaseAnalytics 10.21.0
* FirebaseCrashlytics 10.21.0
* FirebaseFirestore 10.21.0

### Bug Report

I get  EXC_BAD_ACCESS (code=1, address=0x10) when add FLEX to my project (swift with UIKit).
I use Cocoapods 1.15.2, static libraries 
`use_frameworks! :linkage => :static` in `Podfile` 

Use in code:
```
import FLEX

...

        navigationItem.rightBarButtonItem = UIBarButtonItem(title: "FLEX",
                                                            style: .plain,
                                                            target: self,
                                                            action: #selector(flexButtonAction(_:)))
...

    @objc func flexButtonAction(_ sender: UIBarButtonItem) {
        FLEXManager.shared.toggleExplorer()
    }
```

1. static libraries 

https://imgur.com/Hq3ZMMg

```
Thread 0 Crashed::  Dispatch queue: com.apple.main-thread
0   My Project                 	       0x1104d1a7a ___logos_method$_ungrouped$FIRDocumentReference$setData$merge$completion$_block_invoke + 117 (FLEXNetworkObserver.m:288)
1   My Project                 	       0x1105f7d46 firebase::firestore::util::MakeCallback(void (NSError*) block_pointer)::$_0::operator()(firebase::firestore::util::Status) const + 47 (error_apple.mm:60) [inlined]
2   My Project                 	       0x1105f7d46 decltype(std::declval<firebase::firestore::util::MakeCallback(void (NSError*) block_pointer)::$_0&>()(std::declval<firebase::firestore::util::Status>())) std::__1::__invoke[abi:v160006]<firebase::firestore::util::MakeCallback(void (NSError*) block_pointer)::$_0&, firebase::firestore::util::Status>(firebase::firestore::util::MakeCallback(void (NSError*) block_pointer)::$_0&, firebase::firestore::util::Status&&) + 81 (invoke.h:394) [inlined]
[...]

PR #692, opened 2024-02-12
[B, 691] fix EXC_BAD_ACCESS in FLEXNetworkObserver

programming086:
Fix bug CRASH EXC_BAD_ACCESS  #691


NSExceptional (approved):
Good catch! Thank you!

PR #694, opened 2024-03-18
Add condition to Selector call

sng1996:
When I created the `URLSessionDelegate` implementation, app crashed because I didn't implement some optional protocol methods and `slf` doesn't respond to those method selectors.

For example, for protocol `URLSessionDataDelegate` I didn't implement function
`optional func urlSession(_ session: URLSession, dataTask: URLSessionDataTask, didReceive data: Data)`

### DIFF
--- a/Classes/Network/PonyDebugger/FLEXNetworkObserver.m
+++ b/Classes/Network/PonyDebugger/FLEXNetworkObserver.m
@@ -589,9 +589,11 @@ static FIRDocumentReference * _logos_method$_ungrouped$FIRCollectionReference$ad
 
         void (^swizzleBlock)(NSURLConnection *) = ^(NSURLConnection *slf) {
             [FLEXNetworkObserver.sharedObserver connectionWillCancel:slf];
-            ((void(*)(id, SEL))objc_msgSend)(
-                slf, swizzledSelector
-            );
+            if ([slf respondsToSelector:swizzledSelector]) {
+                ((void(*)(id, SEL))objc_msgSend)(
+                    slf, swizzledSelector
+                );
+            }
         };
 
         IMP implementation = imp_implementationWithBlock(swizzleBlock);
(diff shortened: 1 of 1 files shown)


Classes/Network/PonyDebugger/FLEXNetworkObserver.m
```diff
@@ -589,9 +589,11 @@ + (void)injectIntoNSURLConnectionCancel {
...
         void (^swizzleBlock)(NSURLConnection *) = ^(NSURLConnection *slf) {
             [FLEXNetworkObserver.sharedObserver connectionWillCancel:slf];
-            ((void(*)(id, SEL))objc_msgSend)(
-                slf, swizzledSelector
-            );
+            if ([slf respondsToSelector:swizzledSelector]) {
+                ((void(*)(id, SEL))objc_msgSend)(
+                    slf, swizzledSelector
+                );
+            }
```
NSExceptional, line 596: I don't think this check is necessary, since `NSURLConnection` will always implement `-cancel`. Do you agree?

Classes/Network/PonyDebugger/FLEXNetworkObserver.m
```diff
@@ -680,10 +682,11 @@ + (void)swizzleResumeSelector:(SEL)selector forClass:(Class)class {
...
         }
-
-        ((void(*)(id, SEL))objc_msgSend)(
-            slf, swizzledSelector
-        );
+        if ([slf respondsToSelector:swizzledSelector]) {
+            ((void(*)(id, SEL))objc_msgSend)(
+                slf, swizzledSelector
+            );
+        }
```
NSExceptional, line 689: Same here

Classes/Network/PonyDebugger/FLEXNetworkObserver.m
```diff
@@ -745,13 +748,17 @@ typedef void (^SendAsyncRequestBlock)(
...
+                }
             } else {
-                ((void(*)(id, SEL, id, id, id))objc_msgSend)(
-                    slf, swizzledSelector, request, queue, completion
-                );
+                if ([slf respondsToSelector:swizzledSelector]) {
+                    ((void(*)(id, SEL, id, id, id))objc_msgSend)(
+                        slf, swizzledSelector, request, queue, completion
+                    );
+                }
```
NSExceptional, line 761: These as well, if I am reading correctly and these are on `NSURLConnection`

NSExceptional:
I stopped reviewing the PR when I realized just how many of these changes you made.

Can you add the check in the one place you need it instead of everywhere? There is an overhead to using `respondsToSelector:`, so we should only use it when we know the method might be optional, not in cases where we are swizzling methods that are guaranteed to exist.

NSExceptional (changes requested):
I stopped reviewing the PR when I realized just how many of these changes you made.

Can you add the check in the one place you need it instead of everywhere? There is an overhead to using `respondsToSelector:`, so we should only use it when we know the method might be optional, not in cases where we are swizzling methods that are guaranteed to exist.

PR #704, opened 2024-07-16
FIx toolbar visibility bug

ataetgi:
Sometimes the toolbar gets hidden and never becomes visible again.

### DIFF
--- a/Classes/Core/Controllers/FLEXNavigationController.m
+++ b/Classes/Core/Controllers/FLEXNavigationController.m
@@ -209,6 +209,8 @@
             [self setToolbarHidden:NO animated:YES];
         } else if (yTranslation < -20) {
             [self setToolbarHidden:YES animated:YES];
+        } else {
+            [self setToolbarHidden:NO animated:YES];
         }
     }
 }
--- a/Classes/ViewHierarchy/SnapshotExplorer/FHSView.m
+++ b/Classes/ViewHierarchy/SnapshotExplorer/FHSView.m
@@ -88,6 +88,10 @@
 @implementation FHSView (Snapshotting)
 
 + (UIImage *)drawView:(UIView *)view {
+    if (CGRectIsEmpty(view.bounds)) {
+        return [UIImage new];
+    }
+
     CGSize size = view.bounds.size;
     CGRect bounds = view.bounds;
     CGFloat minUnit = 1.f / UIScreen.mainScreen.scale;


ataetgi:
@NSExceptional fyi

NSExceptional:
Do we know what is causing the bug here exactly?

FYI you can tap the navigation bar title to make it reappear

ataetgi:
> Do we know what is causing the bug here exactly?
> 
> FYI you can tap the navigation bar title to make it reappear

When scrolling down, the toolbar hides. However, when I scroll back up, it doesn’t enter any if block and remains hidden.

ataetgi:
Before:
https://github.com/user-attachments/assets/0e852c58-4d51-46cd-820f-a0d74cf4f48c

After:
https://github.com/user-attachments/assets/3930e551-ee16-43b2-acda-78400d59c58b

@NSExceptional

NSExceptional:
> Before:
> 
> https://github.com/user-attachments/assets/0e852c58-4d51-46cd-820f-a0d74cf4f48c

You have to scroll up faster to make it appear. The conditions are `self.canShowToolbar`, `yTranslation > 20`, and `velocity > 250`

Can you try flicking it up instead of just dragging up casually? This is by design so that I personally don't become over stimulated by the toolbar constantly disappearing and reappearing.

ataetgi:
> > Before:
> > 
> > https://github.com/user-attachments/assets/0e852c58-4d51-46cd-820f-a0d74cf4f48c
> 
> You have to scroll up faster to make it appear. The conditions are `self.canShowToolbar`, `yTranslation > 20`, and `velocity > 250`
> 
> Can you try flicking it up instead of just dragging up casually? This is by design so that I personally don't become over stimulated by the toolbar constantly disappearing and reappearing.

@NSExceptional I have tried many times but its very hard to trigger to show toolbar :( especially you if don’t know the exact code to trigger like testers in our project

NSExceptional:
I see

Well, I cannot merge this as-is, because it basically says "if all other logic fails, show the toolbar" which is the opposite of what we want here

Maybe try adjusting the velocity required to show the toolbar again?

I would also accept a PR that adds a preprocessor directive that disables toolbar hiding entirely

ataetgi:
> I see
> 
> Well, I cannot merge this as-is, because it basically says "if all other logic fails, show the toolbar" which is the opposite of what we want here
> 
> Maybe try adjusting the velocity required to show the toolbar again?
> 
> I would also accept a PR that adds a preprocessor directive that disables toolbar hiding entirely

I investigated a bit more and noticed that in the simulator, yVelocity is coming as 0. I’m not sure if it’s due to my project.”

PR #711, opened 2025-02-21
Add option to ignore specific first responder classes for keyboard shortcuts

PoeticPete:
There are instances where a view will be first responder, but we still want to allow keyboard shortcuts.

This change adds an ignore-list for allowing FLEX keyboard shortcuts even if there is a first responder on the given classes.

### DIFF
--- a/Classes/Manager/FLEXManager+Extensibility.h
+++ b/Classes/Manager/FLEXManager+Extensibility.h
@@ -78,6 +78,11 @@ NS_ASSUME_NONNULL_BEGIN
                                   action:(dispatch_block_t)action
                              description:(NSString *)description;
 
+/// The simulator shortcuts will no-op if there is already a first responder in the window.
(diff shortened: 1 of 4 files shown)


PoeticPete:
> Why the `simulator` specific language? I don't think these are specific to the simulator, are they? Do they not also work when you have a bluetooth keyboard connected to the device?
> 
> 
> 
> (If they _are_ simulator-only, we need to add some `///` comments to this class to reflect that)

This logic indeed simulator specific. See `registerSimulatorShortcutWithKey ` for more context.

The comments above the new methods already mention simulator. The entire Classes/Utility/Keyboard/FLEXKeyboardShortcutManager.m is already gated to simulators only. Is the request to add comments at the top of FLEXKeyboardShortcutManager.h? That feels a little beyond the scope of this PR, but I'd be happy to publish a new one addressing that.

PR #712, opened 2025-03-06
Remove [UIApplication openURL:]

dmaclach:
[UIApplication openURL:] has been deprecated since iOS 10, and has been rendered useless if compiling with Xcode 16. Since the earliest supported OS by FLEX is 10.3, let's remove this method from FLEX so that it doesn't trip static analyzers looking for the dead method.


matrush (approved):
lgtm

Issue #716, opened 2025-06-02
SpringBoard Crash when loading ViewHierarchy SnapshotExplorer on iOS 18.2, iPhone 16


EthanArbuckle:
### Environment
- iOS 18.2 - iPhone 16 (Simulator)
- FLEX @ 2bfba6715eff664ef84a02e8eb0ad9b5a609c684

### Bug Report

Attempting to enter the View Hierarchy Snapshot Explorer consistently leads to a SpringBoard crash on iOS 18.2 (simulator) when using an iPhone 16 device.

This crash does not occur on every iOS 18.2 device type

* iPhone 16 (iOS 18.2 sim): Crashes
* iPad Pro M4 (iOS 18.2 sim): Does not crash

(I did not test to see if other device models were affected)

Crash reports:

https://ghostbin.cloud/utazy
https://ghostbin.cloud/b5hvy
https://ghostbin.cloud/chb7k

Relevant logs:

```
ReportCrash:759 ASI found [CoreFoundation] (sensitive) '*** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: '-setHidden: is unavailable on SBHomeGrabberView''
ReportCrash:759 ASI found [libc++abi.dylib] (sensitive) 'terminating due to uncaught exception of type NSException'
ReportCrash:759 ASI found [libsystem_c.dylib] (sensitive) 'abort() called'
ReportCrash:759 ASI found [libsystem_sim_platform.dylib] (sensitive) 'CoreSimulator 993.7 - Device: iPhone 16 (D134598B-4CBD-4BB8-A2E4-FB86BD7758CB) - Runtime: iOS 18.2 (22C150) - DeviceType: iPhone 16'
```

Triggering the crash:
 
![Image](https://github.com/user-attachments/assets/b4623bda-97a8-4d7e-aaa9-b90efd608fd2)

---

Adding a `respondsToSelector:` check here will probably fix it: https://github.com/FLEXTool/FLEX/blob/2bfba6715eff664ef84a02e8eb0ad9b5a609c684/Classes/ViewHierarchy/SnapshotExplorer/FHSView.m#L143

edit: well that is annoying:

```
- (void)setHidden:(BOOL)a3 {
  objc_msgSend(MEMORY[0x1E4F28B00], "currentHandler", a3);
  id v5 = (id)objc_claimAutoreleasedReturnValue();
[...]

PR #718, opened 2025-06-06
fix: snapshot explorer crashes springboard on some iOS 18 devices

EthanArbuckle:
Addressing issue #716: Attempting to enter the View Hierarchy Snapshot Explorer leads to a SpringBoard crash on iOS 18.2 (simulator) when using an iPhone 16 device.

This change introduces a helper method for hiding/unhiding views in `FHSView.m`, `_setHidden:forView:`, which handles potential exceptions raised when interacting with some classes.

For more details about the crash: https://github.com/FLEXTool/FLEX/issues/716

### DIFF
--- a/Classes/ViewHierarchy/SnapshotExplorer/FHSView.m
+++ b/Classes/ViewHierarchy/SnapshotExplorer/FHSView.m
@@ -123,7 +123,7 @@
     }
 
     if (!rootView.isHidden) {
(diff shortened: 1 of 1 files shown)


NSExceptional (approved):

NSExceptional:
Ty ty 🙏🏻

PR #721, opened 2025-06-30
Add `linkerSettings`

brentleyjones:


NSExceptional (approved):
Kinda wonder how we got away with not linking these things before. Thanks!

PR #723, opened 2025-09-11
Force LTR layout for FLEX debugging tool to maintain technical readability

Mahmoud-Abdelwahab:
📋 Summary
This PR ensures that the FLEX debugging tool maintains a consistent Left-to-Right (LTR) layout regardless of the host app's language settings. This is crucial for debugging tools as they display technical content like API requests, responses, and system information that should always be presented in LTR format for optimal readability.

🎯 What's Changed
✅ Force LTR Layout: Implemented semantic layout direction forcing to LTR

### DIFF
--- a/Classes/Core/Controllers/FLEXTableViewController.m
+++ b/Classes/Core/Controllers/FLEXTableViewController.m
@@ -232,6 +232,12 @@ CGFloat const kFLEXDebounceForExpensiveIO = 0.5;
 - (void)viewDidLoad {
     [super viewDidLoad];
     
+    // Force LTR layout for all FLEX table views
+    if (@available(iOS 9.0, *)) {
+        self.view.semanticContentAttribute = UISemanticContentAttributeForceLeftToRight;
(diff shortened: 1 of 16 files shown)


Classes/Core/Controllers/FLEXTableViewController.m
```diff
@@ -232,6 +232,12 @@ - (void)loadView {
 - (void)viewDidLoad {
     [super viewDidLoad];
     
+    // Force LTR layout for all FLEX table views
+    if (@available(iOS 9.0, *)) {
+        self.view.semanticContentAttribute = UISemanticContentAttributeForceLeftToRight;
+        self.tableView.semanticContentAttribute = UISemanticContentAttributeForceLeftToRight;
+    }
+
```
NSExceptional: I _think_ FLEX's minimum supported target is iOS 9, is it not?

NSExceptional (changes requested):
Per my previous comment, I would like to brainstorm a better solution than setting these attributes on each individual view and view controller

PR #702, opened 2024-06-18
Make FLEXFileBrowserController public

kolyuchiy:
This allows to add custom menu item to browse application group container or other path of interest.

This is related to #668.

PR #725, opened 2025-10-31
Silence build warnings when used in Xcode 26 via SwiftPM

Pretz:
Xcode 26 has started treating warnings in SPM headers as project-level warnings for users of the package. This causes issues for any project that has warnings as errors enabled and imports FLEX.

See similar discussions at
https://github.com/facebook/facebook-ios-sdk/issues/2602 and https://github.com/RevenueCat/purchases-ios/issues/5290

Fix two issues with FLEX that hit this change:
- Wrap uses of `UIScreen.mainScreen` in FLEXMacros.h in `clang diagnostic ignored "-Wdeprecated-declarations"`
- Add `FLEXFileBrowserController.h` to `FLEX.h` and mark as public in the Xcode project config.  Since `FLEXFileBrowserController.h` is part of the SwiftPM public headers after being added in #702 , it must be declared in the umbrella header to avoid `FLEX/Classes/Headers/FLEX.h:26:1 umbrella header for module 'FLEX' does not include header 'FLEXFileBrowserController.h'`. Once added to the umbrella header, the Xcode project would not build because the header was not marked public in the project file.

### DIFF
--- a/Classes/FLEX.h
+++ b/Classes/FLEX.h
@@ -13,6 +13,7 @@
 
 #import "FLEXExplorerToolbar.h"
 #import "FLEXExplorerToolbarItem.h"
+#import "FLEXFileBrowserController.h"
 #import "FLEXGlobalsEntry.h"
 
 #import "FLEX-Core.h"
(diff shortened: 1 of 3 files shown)


NSExceptional (changes requested):

Pretz:
Looks like these issues have now been addressed! Thanks!

PR #732, opened 2026-02-04
Fix crash in FLEXNetworkObserver class enumeration

dinhvh:
- Replace objc_getClassList with objc_copyClassList for safer iteration
- Add NULL check for class pointers before accessing
- Skip metaclasses since only instance methods need swizzling
- Validate class registration by name lookup before use
- Add NULL check for class_copyMethodList return value

This fixes crashes when classes are dynamically loaded/unloaded during background thread iteration, which can cause stale pointer access.


NSExceptional (approved):

marcel-glabs:
Hi @NSExceptional - any plans to bring this into a release? We're seeing this crash when using Observation framework on iOS 16, even though the actual usage is guarded with availability checks. Thanks in advance!

PR #735, opened 2026-04-01
Add visual tap indicator to FLEX selection

matrush:
Right now when using selection tool, it's hard to know where we just tapped.
This changes shows a brief semi-transparent blue dot (20x20 circle) at the tap point during view selection.

The indicator fades out over 0.3s after a 0.4s delay and is stored as a weak reference to avoid retain cycles.

https://github.com/user-attachments/assets/2ecad4c8-428b-43fa-931f-31d48e5aba1a

### DIFF
--- a/Classes/ExplorerInterface/FLEXExplorerViewController.m
+++ b/Classes/ExplorerInterface/FLEXExplorerViewController.m
@@ -58,6 +58,9 @@ typedef NS_ENUM(NSUInteger, FLEXExplorerMode) {
 /// The view that we're currently highlighting with an overlay and displaying details for.
 @property (nonatomic) UIView *selectedView;
 
+/// Brief visual indicator shown at the tap point, created once and reused.
+@property (nonatomic) UIView *tapIndicatorView;
+
 /// A colored transparent overlay to indicate that the view is selected.
 @property (nonatomic) UIView *selectedViewOverlay;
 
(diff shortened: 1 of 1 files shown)


NSExceptional (changes requested):

Classes/ExplorerInterface/FLEXExplorerViewController.m
```diff
@@ -634,6 +643,19 @@ - (void)handleSelectionTap:(UITapGestureRecognizer *)tapGR {
         // Thanks to @lascorbe for finding this: https://github.com/Flipboard/FLEX/pull/31
         CGPoint tapPointInView = [tapGR locationInView:self.view];
         CGPoint tapPointInWindow = [self.view convertPoint:tapPointInView toView:nil];
+
+        // Show a brief visual indicator at the tap point
+        [self.tapIndicatorView.layer removeAllAnimations];
+        self.tapIndicatorView.center = tapPointInView;
+        self.tapIndicatorView.alpha = 1;
+        self.tapIndicatorView.hidden = NO;
+        [self.view addSubview:self.tapIndicatorView];
```
NSExceptional: I think this call needs to be alongside where it is created, I'm not sure what happens when you call addSubview multiple times with the same view without ever removing it

NSExceptional (changes requested):

NSExceptional (approved):

NSExceptional:
Thanks! :)

